PR #5, opened 2021-06-25
Stubbed out control flow

zmanian:
Remaining tasks.

[ ] remove commands/cellarcontract.rs @hannydevelop
[ ] update data model to handle multiple pairs and cellar contracts
[ ] implement the decision function
[ ] remove all compiler warnings @hannydevelop
[ ] convert all cellar wrapper functions to use call like `remove_liquidity_eth_from_uni_v3` @hannydevelop
[ ] complete all todos


hannydevelop:
I have fixed this:

[ ] remove commands/cellarcontract.rs
[ ] convert all cellar wrapper functions to use call like remove_liquidity_eth_from_uni_v3

for the compiler warnings, I turned off `missing docs` warning. I also have turned off `unused_functions` warning in `src/gas.rs`. However, I think it's messy with `#[allow(dead_code)]` in each function. What do you think? If it's OK then I'll implement the same in `collector.rs`, `time_range.rs`, and `collector/poller.rs` that still have warnings

PR #9, opened 2021-08-30
contract monitor docker container

levicook:


levicook:
Thinking this will be a useful building block to have when it's time to compose e2e tests.

PR #12, opened 2021-09-01
adding orchestrator in cargo.toml

hannydevelop:


src/commands.rs
```diff
@@ -64,6 +66,10 @@ pub enum CellarRebalancerCmd {
 
     #[options(help = "remove_funds")]
     RemoveFunds(RemoveFundsCmd),
+
+    // Gorc subcommand from gravity-bridge
+    #[options(help = "gorc sub-commands")]
+    Orchestrator(OrchestratorCmd),
```
tony-iqlusion: For now, would it be possible to make `OrchestratorCmd` a thin wrapper around something shared in another crate, and copy-and-paste its definition into this application?

Otherwise we can look at trying to make it generic around the underlying Abscissa application, but that's a lot more work.

zmanian:
@hannydevelop did you test if `print-config` still works? otherwise this looks ready to merge.

hannydevelop:
Hello @zmanian `cargo check --all && cargo test --all` did not throw any error, I'll confirm again. Before merging can you check this out https://github.com/PeggyJV/gravity-bridge/pull/181.

PR #13, opened 2021-09-14
working on cosmos mode

hannydevelop:


src/commands/cosmos_mode.rs
```diff
@@ -0,0 +1,133 @@
...
+            let client = SignerMiddleware::new(client, wallet.clone());
+            let client = Arc::new(client);
+            let mongo = config.mongo.clone();
+
+            let name = &config
+            .keys
+            .rebalancer_key;
+            let cosmos_key = config.load_deep_space_key(name.clone());
+
+            let poller = Poller::new(&cellar, client, &mongo, &cosmos_key)
```
zmanian: So I think hen the poller gets constructed an argument should be passed in determining if it's in single signer or cosmos mode

src/collector/poller.rs
```diff
@@ -109,6 +149,28 @@ impl<T: 'static + Middleware> Poller<T> {
             Ok(())
         } else {
             tick_info.reverse();
+            // TODO(Levi) needs to be initialized (should it be derived from the cosmos_key??)
+            let delegate_cosmos_address =
```
zmanian: Should be derived from the cosmos key that gets passed to the poller constructor

zmanian (changes requested):

src/collector/poller.rs
hannydevelop: Done https://github.com/PeggyJV/cellar_rebalancer_rs/blob/c3f93f33563adcd63ff450bf84384f464b67a8ec/src/collector/poller.rs#L155

hannydevelop:
I'll have to start this PR over and close up this, that's my goal today.

PR #23, opened 2021-10-13
Update Cellars to the latest Abscissa with Clapv3

hannydevelop:


src/commands.rs
```diff
@@ -26,91 +26,91 @@ mod sign_delegate_keys;
...
+    fn run(&self) {
+        self.cmd.run()
+    }
+}
+
 /// This trait allows you to define how application configuration is loaded.
-impl Configurable<CellarRebalancerConfig> for CellarRebalancerCmd {
+impl Configurable<CellarRebalancerConfig> for EntryPoint {
     /// Location of the configuration file
     fn config_path(&self) -> Option<PathBuf> {
```
zmanian, line 115: I think you need to make a similar change to the config_path to what I did here
https://github.com/PeggyJV/gravity-bridge/pull/222/

hannydevelop: Ok!

zmanian (approved):

PR #24, opened 2021-10-13
setValidator wrapper method and command

cbrit:


src/cellar_wrapper.rs
```diff
@@ -149,6 +149,24 @@ impl<T: 'static + Middleware> CellarState<T> {
...
+        let mut call = self.contract.set_validator(validator, value);
+
+        if let Some(gas_price) = self.gas_price {
+            call = call.gas_price(gas_price)
+        }
+
+        let gased = call.gas(5_000_000);
+
+        let pending = gased.send().await?;
+        dbg!(&pending);
```
zmanian, line 168: If we decide to use Infura, we start await the pending tx instead of this dbg! patten.

zmanian (approved):

PR #31, opened 2021-11-10
Ukpai/rust workflow

hannydevelop:


src/commands/cosmos_to_eth.rs
```diff
@@ -47,8 +47,6 @@ impl Runnable for CosmosToEthCmd {
         let gravity_denom = gravity_denom.to_string();
         let is_cosmos_originated = !gravity_denom.starts_with("gravity");
 
-        let gas_limit = config.cosmos.gas_limit;
```
EricBolten, line 50: After the `send_request_batch_tx` fix I don't think `gas_limit` is used anywhere any more. Should we remove the rest of the references to it?

EricBolten: Along those same lines, are any external folks using the rebalancer yet? Wondering if the changes to the config could have any impact on people running the rebalancer currently.

hannydevelop: I removed it after the `send_request_batch_tx` fix. It's no longer there.

The rebalancer is supposed to be run on app.sommelier.finance. However, I don't think it is live yet, seeing that we haven't finished with the double signer mode.

EricBolten: It's still in a couple places, just not used by anything:

https://github.com/PeggyJV/cellar_rebalancer_rs/blob/73ba1f8fd9601e51bbbab99383c1e6848373ffaf/src/config.rs#L193

https://github.com/PeggyJV/cellar_rebalancer_rs/blob/73ba1f8fd9601e51bbbab99383c1e6848373ffaf/src/commands/tx/cosmos.rs#L132

This last one actually seems like a similar issue where the types are the same. The current version of the orchestrator's main loop also expects `gas_adjustment`.

https://github.com/PeggyJV/cellar_rebalancer_rs/blob/73ba1f8fd9601e51bbbab99383c1e6848373ffaf/src/commands/orchestrator/start.rs#L97

https://github.com/PeggyJV/gravity-bridge/blob/d2abe0b5fa9a70a51233d2ffab77d7aa1fcb5c3c/orchestrator/orchestrator/src/main_loop.rs#L59

EricBolten (approved):
LGTM

PR #33, opened 2021-11-23
create bindings for ABI's

hannydevelop:


src/cellar_uniswap_wrapper.rs
```diff
@@ -32,9 +26,16 @@ impl<T: 'static + Middleware> UniswapV3CellarState<T> {
     }
 
     // Rebalance portfolio with cellar tick info
-    pub async fn rebalance(&mut self, cellar_tick_info: Vec<UniswapV3CellarTickInfo>) -> Result<(), Error> {
-        let mut ticks: Vec<(U256, i32, i32, u32)> =
-            cellar_tick_info.into_iter().map(|x| x.to_tuple()).collect();
+    pub async fn rebalance(&mut self, cellar_tick_info: Vec<CellarTickInfo>) -> Result<(), Error> {
+        let mut ticks: Vec<CellarTickInfo> = cellar_tick_info
```
EricBolten: After removing the wrapper, the argument is already of the type CellarTickInfo, so you could just make a mutable clone instead.

src/collector/poller.rs
```diff
@@ -25,6 +26,17 @@ pub struct Poller<T: Middleware> {
     pool: PoolState<T>,
 }
 
+pub fn from_tick_weight(
+    tick_weight: &crate::time_range::TickWeight,
```
EricBolten: You can add a `use` statement to import TickWeight so this is a bit cleaner.

src/commands/fund_cellar.rs
```diff
@@ -26,6 +27,24 @@ pub struct FundCellarCmd {
     pub amount_1: f64,
 }
 
+pub fn cellar_add_params(
```
EricBolten: I don't think these functions are necessary since none of the values are mutated, the caller could just declare a `CellarAddParams` directly.

src/commands/remove_funds.rs
```diff
@@ -8,15 +8,32 @@ use num_traits::Zero;
...
     prelude::*,
     uniswap_pool::PoolState,
 };
+use rebalancer_abi::cellar_uniswap::*;
 
 /// Remove funds from Cellars
 #[derive(Command, Debug, Clap)]
 pub struct RemoveFundsCmd {}
 
+pub fn cellar_remove_params(
```
EricBolten: Same here, the caller could directly instantiate the struct.

EricBolten (commented):
Added a few suggestions.

src/commands/remove_funds.rs
hannydevelop: Fixed with this commit: https://github.com/PeggyJV/cellar_rebalancer_rs/pull/33/commits/8f4daaad16ad94376c5ef59c50719b0a22b34c9f

src/commands/fund_cellar.rs
hannydevelop: Fixed with this commit: https://github.com/PeggyJV/cellar_rebalancer_rs/pull/33/commits/508238cd7ae590059d61a6bbca992e333508e4b2

src/collector/poller.rs
hannydevelop: Fixed with this commit: https://github.com/PeggyJV/cellar_rebalancer_rs/pull/33/commits/ab548a39d77faf9ef220e4f47a54e2751c64cc04

EricBolten:
The only other thing I'd recommend is updating `.gitignore` to ignore the `target` directory when you `cargo run` the `abi_build` crate. This can be done by either removing the leading `/` from `/target` or adding `abi_build/target`.

EricBolten (approved):
LGTM

PR #35, opened 2021-12-02
Ukpai/double signer

hannydevelop:


cellar_rebalancer/src/collector/poller.rs
```diff
@@ -71,11 +88,61 @@ impl<T: 'static + Middleware> Poller<T> {
...
+                        .map(|tick_weight| proto::TickRange {
+                            upper: tick_weight.upper_bound as i32,
+                            lower: tick_weight.lower_bound as i32,
+                            weight: tick_weight.weight as u32,
+                        })
+                        .collect(),
+                }),
+                current_price: self.contract_state.gas_price.unwrap().as_u64(),
+            }),
+            salt: "".to_string(), //TODO: Add salt,
```
EricBolten, line 126: I think this is the last important thing we should be sure to update before merging. Not sure there's anything in the spec as to how large this should be but generating a random 8 character string using the OS randomness should be sufficient I'd imagine.

cellar_rebalancer/src/collector/poller.rs
```diff
@@ -117,10 +212,78 @@ impl<T: 'static + Middleware> Poller<T> {
...
+                    println!("Couldn't Send Precommit {:?}", e);
+                }
+            }
+
+            match timeout(Duration::from_secs(100), async {
+                loop {
+                    // Checking Pre-commits for validators
+                    let res = somm_send::query_allocation_precommits(grpc_client).await;
+                    if let Ok(val) = res {
+                        break val;
```
EricBolten, line 264: Ideally we delay for at least a block here before breaking and proceeding to commit, but this isn't a blocker for merge.

cbrit (approved):

cellar_rebalancer/src/collector/poller.rs
EricBolten: Probably something along the lines of

```
use rand::{distributions::Alphanumeric, rngs::OsRng};

...

salt: OsRng
    .sample_iter(&Alphanumeric)
    .take(8)
    .map(char::from)
    .collect();
```

EricBolten (approved):
Going to merge to unblock parallel development and will follow up to fix any remaining issues in our next set of changes.

Issue #40, opened 2021-12-16
Certificate creation and cert/key storage


cbrit:
With the use of TLS we need a signed server certificate and client CA root cert with which to issue a signed client cert to the Data Science team. The keys and certs need to be stored securely. 

We're leaving the larger process of issuing new certs and onboarding Data Providers to future work.

cbrit:
Currently waiting on the Data Science team to generate and provide client certs

PR #47, opened 2021-12-16
Steward, Poller replacement

### LINKED ISSUE #37: Remove Poller and call Allocation workflow from gRPC handler
The `collector` module needs to be entirely removed. Currently the `poller` is tightly coupled with the allocation workflow (the stuff that calls Sommelier to reach consensus on a rebalance) and the latter needs to broken into its own module and utilized by the cellar handler.

cbrit:
Closes #37

Lots happening here:

1. Eric and I created the proto for the new data provision model, located in `steward_proto` and built via `steward_proto_build`
2. The startup command(s) now configure TLS and start a `tonic` server that implements the steward proto. Currently it's looking for paths in the config file to read in certs. This will probably need to change.
3. `collector`, `poller` and several other legacy modules were removed.
4. The new `allocation` workflow that @hannydevelop implemented now lives in the `allocation.rs` module.
5. The `cellar_uniswap_wrapper.rs` was combined with the proto handling code into a `cellars/uniswapv3` module/submodule.
6. Created a `Dockerfile`. May or may not be working right now.

Largely a WIP, the items listed in [the project ](https://github.com/PeggyJV/cellar_rebalancer_rs/projects/2)still need to be completed.


cellar_rebalancer/src/commands/single_signer.rs
```diff
@@ -1,103 +1,53 @@
...
+
+            // Configure TLS
+            let cert = tokio::fs::read(&config.tls.server_cert).await.unwrap();
+            let key = tokio::fs::read(&config.tls.server_key).await.unwrap();
+            let server_identity = Identity::from_pem(cert, key);
+            let client_ca_cert = tokio::fs::read(&config.tls.client_ca_cert).await.unwrap();
+            let client_ca_cert = Certificate::from_pem(client_ca_cert);
+            let tls_config = ServerTlsConfig::new()
+                .identity(server_identity)
+                .client_ca_root(client_ca_cert);
```
zmanian: Does this api only support a ca_client_root?

it would be good document how to test the server endpoints in a comment or README

cellar_rebalancer/src/cellars/uniswapv3.rs
```diff
@@ -158,11 +166,42 @@ impl<T: 'static + Middleware> UniswapV3CellarState<T> {
...
+        let eth_gas_price = match cellars::get_gas_price().await {
+            Ok(gp) => gp,
+            Err(err) => {
+                error!("Failed to get cellar gas price: {:?}", err);
+                // what's the right way to handle this?
+                0.into()
+            }
+        };
+
+        allocation::decide_rebalance(tick_ranges, pair_id, eth_gas_price.as_u64()).await;
```
zmanian: I wonder if we should await on this or tokio spawn this into a new thread and return.

Errors in the allocation module aren't really the client's responsibility to fix

cellar_rebalancer/src/commands/single_signer.rs
cbrit: This might be in the scope of #40? The current state of the cert config was just where we left it to get a working example but I think it definitely need fleshing out.

cellar_rebalancer/src/cellars/uniswapv3.rs
cbrit: Pushed a fix

cellar_rebalancer/src/commands/single_signer.rs
EricBolten: This works as an MVP as long as there is only one data provider. To support more, you'd need to use `rustls_server_config` rather than `client_ca_root` to use a handcrafted config that supports multiple roots of trust.

Cargo.toml
```diff
@@ -6,13 +6,23 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [workspace]
-default-members = ["abi_build", "cellar_rebalancer", "rebalancer_abi"]
+default-members = [
+	"cellar_rebalancer",
+    "rebalancer_abi",
```
EricBolten: Nit: indentation.

Cargo.toml
```diff
@@ -6,13 +6,23 @@ edition = "2021"
...
+	"steward_proto",
+	"steward_proto_build",
 ]
 [dependencies]
-abi_build = {path = "./abi_build"}
-cellar_rebalancer = {path = "./cellar_rebalancer"}
-rebalancer_abi = {path = "./rebalancer_abi"}
\ No newline at end of file
+cellar_rebalancer = { path = "./cellar_rebalancer" }
+rebalancer_abi = { path = "./rebalancer_abi" }
```
EricBolten: We probably want the ABI crate and the ABI build crate to have the same naming scheme, otherwise it may be confusing.

cellar_rebalancer/src/cellars/uniswapv3.rs
```diff
@@ -158,11 +166,44 @@ impl<T: 'static + Middleware> UniswapV3CellarState<T> {
...
-// Struct for UniswapV3CellarTickInfo
-#[derive(Clone, Debug)]
-pub struct UniswapV3CellarTickInfo {
-    pub(crate) token_id: U256,
-    pub(crate) tick_upper: i32,
-    pub(crate) tick_lower: i32,
-    pub(crate) weight: u32,
+pub struct UniswapV3CellarAllocator;
+
+#[async_trait]
```
EricBolten, line 171: This is the only import from `axum` used in the project. Is it necessary? I believe `tonic` also reexports this.

cellar_rebalancer/src/allocation.rs
```diff
@@ -0,0 +1,157 @@
+use crate::{cellars::uniswapv3::UniswapV3CellarTickInfo, prelude::*, somm_send};
+use abscissa_core::Application;
+use deep_space::{Coin, Contact};
+use ethers::prelude::U256;
+use proto::query_client::QueryClient as AllocationQueryClient;
+use somm_proto::somm as proto;
```
EricBolten: This layer of indirection is a bit confusing, and we also have other proto definitions in the project.

cellar_rebalancer/src/allocation.rs
```diff
@@ -0,0 +1,157 @@
...
+use std::time::Duration;
+use tokio::time::{sleep, timeout};
+use tonic::transport::Channel;
+
+pub struct Connections {
```
EricBolten, line 11: We should try to resolve whether we successfully created clients or not at the time of instantiating them, this way we can pass a `Connections` struct of clients rather than `Options`.

cellar_rebalancer/src/allocation.rs
```diff
@@ -0,0 +1,157 @@
...
+    pub grpc: Option<AllocationQueryClient<Channel>>,
+    pub contact: Option<Contact>,
+}
+
+pub async fn allocation_precommit(
+    cosmos_key: &deep_space::private_key::PrivateKey,
```
EricBolten: Let's add `PrivateKey` to our `deep_space` import above.

cellar_rebalancer/src/allocation.rs
```diff
@@ -0,0 +1,157 @@
...
+    let hasher = somm_send::data_hash(allocation, delegate_cosmos_address)
+        .await
+        .unwrap();
+    proto::AllocationPrecommit {
+        hash: hasher.hash,
+        cellar_id: pair_id,
+    }
+}
+
+pub async fn cellar_query_client() -> Connections {
```
EricBolten: This provides both the `AllocationQueryClient` and the `Contact` for sending Cosmos transactions, so a more generic name might be appropriate.

cellar_rebalancer/src/allocation.rs
```diff
@@ -0,0 +1,157 @@
...
+                // Sending Pre-commits
+                somm_send::send_precommit(
+                    &contact,
+                    delegate_cosmos_address,
+                    cosmos_key,
+                    fee.clone(),
+                    vec![allocation_precommit(&cosmos_key, &allocation, pair_id.to_string()).await],
+                )
+                .await
+                .unwrap();
```
EricBolten, line 107: `decide_rebalance` should return a `Result` and we should propagate errors here rather than unwrapping.

cellar_rebalancer/src/allocation.rs
```diff
@@ -0,0 +1,157 @@
...
+            Err(e) => {
+                println!("Couldn't Send Precommit {:?}", e);
+            }
+        }
+
+        match timeout(Duration::from_secs(100), async {
+            loop {
+                // Checking Pre-commits for validators
+                let res = somm_send::query_allocation_precommits(grpc_client).await;
+                if let Ok(val) = res {
```
EricBolten, line 118: Let's add a one-block delay here before proceeding to `break`, so we submit our commit in the block after our precommit.

cellar_rebalancer/src/allocation.rs
```diff
@@ -0,0 +1,157 @@
...
+                // Sending Commits
+                somm_send::send_allocation(
+                    &contact,
+                    delegate_cosmos_address,
+                    cosmos_key,
+                    fee,
+                    vec![allocation],
+                )
+                .await
+                .unwrap();
```
EricBolten, line 136: Same deal here, let's propagate an error if we hit one.

cellar_rebalancer/src/allocation.rs
```diff
@@ -0,0 +1,157 @@
...
+) -> proto::Allocation {
+    proto::Allocation {
+        vote: Some(proto::RebalanceVote {
+            cellar: Some(proto::Cellar {
+                id: pair_id,
+                tick_ranges: tick_ranges,
+            }),
+            current_price: eth_gas_price,
+        }),
+        salt: "".to_string(), //TODO: Add salt,
```
EricBolten, line 158: We still need to implement this. I included a suggestion in the previously merged PR:

https://github.com/PeggyJV/cellar_rebalancer_rs/pull/35#discussion_r770039091

cellar_rebalancer/src/cellars.rs
```diff
@@ -0,0 +1,19 @@
...
+pub(crate) mod uniswapv3;
+
+const UNISWAPV3_CELLAR: &str = "uniswapv3";
+
+pub async fn get_gas_price() -> Result<U256, Error> {
+    CellarGas::etherscan_standard().await.map_err(|e| e.into())
+}
+
+// returns (chain name, cellar address)
+fn parse_cellar_id(cellar_id: &str) -> (&str, ethers::types::H160) {
```
EricBolten: Should we be using a struct to represent a cellar rather than a tuple?

cellar_rebalancer/src/cellars/uniswapv3.rs
```diff
@@ -158,11 +166,44 @@ impl<T: 'static + Middleware> UniswapV3CellarState<T> {
...
+                lower: d.lower_price,
+                weight: d.weight,
+            })
+            .collect();
+        let pair_id = APP
+            .config()
+            .cellars
+            // sketchy
+            .get(0)
+            .unwrap()
```
EricBolten, line 193: If there is an error here, we should return one by setting our return `tonic::Status` appropriately.

cellar_rebalancer/src/cellars/uniswapv3.rs
```diff
@@ -158,11 +166,44 @@ impl<T: 'static + Middleware> UniswapV3CellarState<T> {
...
+            .pair_id;
+        let eth_gas_price = match cellars::get_gas_price().await {
+            Ok(gp) => gp,
+            Err(err) => {
+                error!("Failed to get cellar gas price: {:?}", err);
+                // what's the right way to handle this?
```
EricBolten: Perhaps this belongs in `decide_rebalance` rather than the request handler. Additionally, you could add some fallbacks to Etherscan if you wanted to: `ethers` can look at price history in the chain to estimate gas price. Another alternative would be to have a config option for a default gas price.

cellar_rebalancer/src/commands/cosmos_mode.rs
```diff
@@ -2,111 +2,56 @@
...
-            let mut tasks = vec![];
-            for poller in pollers {
-                let collector = collector.clone();
-                let task = tokio::spawn(async move { poller.run(collector).await });
-                tasks.push(task);
-            }
-            future::join_all(tasks).await;
+            // Reflection required for certain clients to function... such as grpcurl
+            let contents = fs::read("steward_proto/src/prost/descriptor.bin").unwrap();
+            let proto_descriptor_service = tonic_reflection::server::Builder::configure()
```
EricBolten, line 27: We don't need to enable the reflection service in production. We probably want this behind a debug flag or something.

cellar_rebalancer/src/commands/cosmos_mode.rs
```diff
@@ -2,111 +2,56 @@
...
+                .register_encoded_file_descriptor_set(contents.as_slice())
+                .build()
+                .unwrap();
+
+            // Configure TLS
+            let cert = tokio::fs::read(&config.tls.server_cert).await.unwrap();
```
EricBolten: In general, throwing a `panic` is kind of fine in the beginning of a CLI call, but providing a clean error string that explains the issue back to the end user can be a nicer UX.

cellar_rebalancer/src/commands/cosmos_mode.rs
```diff
@@ -2,111 +2,56 @@
...
+            let key = tokio::fs::read(&config.tls.server_key).await.unwrap();
+            let server_identity = Identity::from_pem(cert, key);
+            let client_ca_cert = tokio::fs::read(&config.tls.client_ca_cert).await.unwrap();
+            let client_ca_cert = Certificate::from_pem(client_ca_cert);
+            let tls_config = ServerTlsConfig::new()
+                .identity(server_identity)
+                .client_ca_root(client_ca_cert);
+
+            // run it
+            let addr = ([127, 0, 0, 1], 3000).into();
```
EricBolten: We may want to adjust this port assignment, it's the default port for a lot of webservers in test. We could also make it a config option but we'd have to ensure data providers would be connecting to a full URL including the port.

cellar_rebalancer/src/commands/single_signer.rs
```diff
@@ -1,103 +1,55 @@
...
-            }
-            future::join_all(tasks).await;
+            // Reflection required for certain clients to function... such as grpcurl
+            let contents = fs::read("steward_proto/src/prost/descriptor.bin").unwrap();
+            let proto_descriptor_service = tonic_reflection::server::Builder::configure()
+                .register_encoded_file_descriptor_set(contents.as_slice())
+                .build()
+                .unwrap();
+
+            // Configure TLS
```
EricBolten, line 27: Since we're repeating boilerplate like this in multiple commands, we should probably move it to a utility function so changes don't have to happen in multiple places if it's updated.

cellar_rebalancer/src/config.rs
```diff
@@ -65,6 +67,24 @@ impl Default for CellarRebalancerConfig {
         }
     }
 }
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct TlsSection {
+    pub server_cert: String,
```
EricBolten: Nit: prefer to suffix strings meant to be paths with `_path` to indicate it's not a serialization or something.

cellar_rebalancer/src/commands/single_signer.rs
```diff
@@ -1,103 +1,55 @@
...
+
+            // run it
+            let addr = ([127, 0, 0, 1], 3000).into();
+            info!("listening on {}", addr);
+            tonic::transport::Server::builder()
+                .tls_config(tls_config)
+                .unwrap_or_else(|err| {
+                    panic!("{:?}", err);
+                })
+                .add_service(UniswapV3CellarAllocatorServer::new(
```
EricBolten: This isn't actually the single signer mode if it's just connecting back to the cosmos mode handler. If you want a single signer command you'd need to provide an alternative that uses the old direct Ethereum methodology.

EricBolten (commented):
Exactly 20 comments, as requested ;)

cellar_rebalancer/src/cellars/uniswapv3.rs
EricBolten: As a counterpoint, clients of most HTTP services can't really do anything about getting 500-level error codes back but it is useful information to know that the endpoint is broken.

Cargo.toml
cbrit: yeah why'd you name it like that ;)

cellar_rebalancer/src/commands/cosmos_mode.rs
cbrit: There is an error handling issue in the project backlog that I'd prefer to leave these fixes to if that's alright.

cellar_rebalancer/src/commands/single_signer.rs
cbrit: You're right, I just pasted this to get it to compile because the previous version was also identical to the cosmos_mode command. There is an issue to fix single-signer mode

cellar_rebalancer/src/allocation.rs
cbrit: Was going to leave this for the error handling issue in the project backlog.

cellar_rebalancer/src/cellars/uniswapv3.rs
EricBolten: Whoops, I just realized, cosmos mode needs cosmos gas prices and not Ethereum gas prices!

cellar_rebalancer/src/cellars/uniswapv3.rs
cbrit: The directive is necessary but you are correct, we can remove the `axum` dep.

cellar_rebalancer/src/cellars/uniswapv3.rs
zmanian: I think it's useful to be able to return an error if the data doesn't parse or is otherwise invalid.

cellar_rebalancer/src/commands/single_signer.rs
cbrit: Removed the handler service just in case

EricBolten (approved):
As discussed, will merge this even though it's still a bit of a WIP. My only request is that we take the remaining action items and fold them into the open tasks so we don't re-review the same code.

PR #50, opened 2021-12-21
Collin/error handling uniswap

### LINKED ISSUE #45: Logging & Error Handling

cbrit:
Closes #45

### DIFF
--- a/steward/src/allocation.rs
+++ b/steward/src/allocation.rs
@@ -28,19 +28,17 @@ pub async fn allocation_precommit(
     cosmos_key: &PrivateKey,
     allocation: &somm::Allocation,
     cellar_address: String,
-) -> somm::AllocationPrecommit {
+) -> Result<somm::AllocationPrecommit, Error> {
     let config = APP.config();
     let delegate_cosmos_address = cosmos_key
-        .to_address(&config.cosmos.prefix)
-        .unwrap()
+        .to_address(&config.cosmos.prefix)?
         .to_string();
     let hasher = somm_send::data_hash(allocation, delegate_cosmos_address)
-        .await
-        .unwrap();
-    somm::AllocationPrecommit {
+        .await?;
+    Ok(somm::AllocationPrecommit {
         hash: hasher.hash,
         cellar_id: cellar_address,
-    }
+    })
 }
 
 pub async fn get_connections() -> Result<Connections, Error> {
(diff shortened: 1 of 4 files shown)


cellar_rebalancer/src/error.rs
```diff
@@ -26,15 +27,24 @@ pub enum ErrorKind {
...
     Http,
+    /// Cryptographic Keys error
+    #[error("key related error")]
+    KeysError,
+    /// Miscellaneous error
+    ///
+    /// Errors that are returned with types that provide no
+    /// categorical information, such as String
+    #[error("allocation error")]
+    MiscError,
```
EricBolten: I don't see any usage of this currently. I don't think this is a useful subtype of error, even if what we're encapsulating is just a string, the context in which it is being returned should provide some useful info for categorizing it. If this is meant to be, as is suggested, an "allocation error", perhaps that is what we mean to call it.

EricBolten (commented):
Small comment re: error typing.

cellar_rebalancer/src/error.rs
zmanian: I don't remember what this was for either

EricBolten:
This will have some merge conflicts to resolve since we've renamed the repo and crate.

cellar_rebalancer/src/error.rs
cbrit: I think this was something I forgot to remove when I realized I didn't need it. There was some function returning a Result<(), String> and I wasn't sure how to handle the String.

cbrit: The method `allocation_precommit` now returns a `Result<somm::AllocationPrecommit, steward::error::Error>`, so each error thrown inside needs to be converted to a `steward::error::Error`. 

On this line, `somm_send::data_hash` result contains an error type `String`, which previously could not be cast as our error type (notice the question mark)

https://github.com/PeggyJV/steward/blob/a90293af91b969f19f3fc7763ec5117234935dbd/steward/src/allocation.rs#L36-L37

How do I approach this more cleanly?

EricBolten: The point I was trying to make here would be more that `data_hash` could be returning an error type that categorizes what type of error it is, rather than just wrapping a string in an `Err` response.

EricBolten (approved):
LGTM. MiscError still present but it's not a blocker, just cruft. Can't merge until you take this out of draft mode though.

EricBolten:
Wait...I think I can just do that myself.

PR #55, opened 2021-12-31
Single signer mode

### LINKED ISSUE #44: Single-Signer mode

hannydevelop:
This closes #44


hannydevelop:
#44 will close this

hannydevelop:
I'm removing this function, it's no longer needed https://github.com/PeggyJV/steward/blob/22dbbaf46a9ab23702a4988dbf2b71fd98a31c9d/steward/src/cellars/uniswapv3.rs#L37

steward_proto/proto/cellars/uniswapv3.proto
```diff
@@ -28,3 +28,15 @@ message RebalanceResponse {}
 //     "check_frequency": 3600,
 //     "schema_type": "UNIv3",
 // }
+
+service UniswapV3DirectCellar {
```
EricBolten: We don't need to implement a proto -- steward is either going to be running in Cosmos voting mode or single signer mode, in either case it can expose the same endpoint and receive the same proto messages as the Cosmos mode, because we do not expect the data provider to adjust their format.

steward/src/allocation.rs
```diff
@@ -181,3 +205,44 @@ pub fn to_allocation(
...
+    let key = key
+        .to_pem()
+        .parse::<k256::elliptic_curve::SecretKey<k256::Secp256k1>>()
+        .expect("Could not parse key");
+
+    let wallet: LocalWallet = Wallet::from(key);
+
+    let eth_host = config.ethereum.rpc.clone();
+    let client = Provider::<Http>::try_from(eth_host.clone()).unwrap();
+    let client = SignerMiddleware::new(client, wallet.clone());
```
EricBolten: In my experience I've run into errors unless I set the chain ID on the wallet. An example from one of the steward commands:

https://github.com/PeggyJV/steward/blob/5948dadad915b4bfa564fd7860dcf5f98d4e9832/steward/src/commands/eth_to_cosmos.rs#L49-L59

steward/src/allocation.rs
```diff
@@ -24,17 +30,31 @@ pub fn bytes_to_hex_str(bytes: &[u8]) -> String {
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct TickWeight {
+    pub upper: i32,
+    pub lower: i32,
+    pub weight: u32,
+}
+
+pub fn from_tick_weight(tick_weight: TickWeight) -> CellarTickInfo {
+    CellarTickInfo {
+        token_id: U256::zero(),
```
EricBolten: Should this be zero when we end up making the contract call? It isn't being updated after the `CellarTickInfo` is created from the incoming `TickWeight` structs.

steward/src/allocation.rs
```diff
@@ -181,3 +205,44 @@ pub fn to_allocation(
...
+    let mut contract_state = UniswapV3CellarState::new(cellar_address, client);
+    for tick_weight in tick_weight {
+        if tick_weight.weight > 0 {
+            tick_info.push(from_tick_weight(tick_weight.clone()))
+        }
+    }
+    if std::env::var("CELLAR_DRY_RUN").expect("Expect CELLAR_DRY_RUN var") == "TRUE" {
+        Ok(())
+    } else {
+        tick_info.reverse();
```
EricBolten: One of the first things that `contract_state.rebalance(tick_info)` does is call `reverse` on `tick_info`. We should remove this call or it will be reversed and then reversed again.

EricBolten (commented):
Some suggestions and questions.

steward/src/allocation.rs
hannydevelop: Fixed with this commit https://github.com/PeggyJV/steward/pull/55/commits/8c72dbd33536f6e42fb57c7bc8d78d03d1dc4dd0

steward/src/allocation.rs
hannydevelop: done! https://github.com/PeggyJV/steward/pull/55/commits/20de02ea777ef1b44f1a7f98e709434beab179ff

steward/src/allocation.rs
hannydevelop: I honestly don't really know, this is what I saw in Zaki's former codes. What do you suggest?

steward_proto/proto/cellars/uniswapv3.proto
hannydevelop: Does this means there shouldn't be a `DirectRebalance` endpoint and a `UniswapV3DirectCellar`? So, I should go ahead with the `UniswapV3CellarAllocator` service?

steward/src/allocation.rs
EricBolten: The update retrieves the chain ID, but you still need to create a cloned wallet associated with that chain ID as such:

```
 let eth_client = SignerMiddleware::new( 
     provider, 
     ethereum_wallet.clone().with_chain_id(chain_id), 
 ); 
```

steward_proto/proto/cellars/uniswapv3.proto
EricBolten: Well, you still want `UniswapV3DirectCellar`. I think you would do this by using `UniswapV3CellarAllocatorServer` when doing `add_service`, but you instead pass it `UniswapV3DirectCellar` which is also an implementation of `server::UniswapV3CellarAllocator` using the `UniswapV3DirectCaller` struct, similar to the cosmos version. This way you are adhering to the same API but can change the internal implementation.

steward/src/allocation.rs
EricBolten: Hrmmm, this depends on how the data is being handled by the target contract's `rebalance` function.  Would have to look into it.

steward_proto/proto/cellars/uniswapv3.proto
hannydevelop: I get an error if I do it this way.

the trait bound `cellars::uniswapv3::UniswapV3DirectCellar: steward_proto::uniswapv3::uniswap_v3_cellar_allocator_server::UniswapV3CellarAllocator` is not satisfied
the trait `steward_proto::uniswapv3::uniswap_v3_cellar_allocator_server::UniswapV3CellarAllocator` is not implemented for `cellars::uniswapv3::UniswapV3DirectCellar`

steward/src/allocation.rs
```diff
@@ -14,6 +23,34 @@ pub struct Connections {
     pub gravity_client: GravityQueryClient<Channel>,
 }
 
+pub fn format_eth_address(address: Address) -> String {
```
EricBolten: Since the address was changed into a String, `format_eth_address` and `bytes_to_hex_str` aren't used any more, so let's delete them.

steward/src/commands/single_signer.rs
```diff
@@ -38,6 +42,9 @@ impl Runnable for SingleSignerCmd {
                 .unwrap_or_else(|err| {
                     panic!("{:?}", err);
                 })
+                .add_service(UniswapV3CellarAllocatorServer::new(
+                    UniswapV3CellarAllocator,
```
EricBolten: You'll want this to be `UniswapV3DirectCellar`.

steward_proto/src/prost/cellars.uniswapv3.v1.rs
```diff
@@ -18,6 +18,17 @@ pub struct RebalanceRequest {
 }
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct RebalanceResponse {}
+#[derive(Clone, PartialEq, ::prost::Message)]
```
EricBolten: We need to re-run the proto build now that we have deleted the new proto, so that this file isn't including functions for a proto that doesn't exist any more (and that will also fix `descriptor.bin`, which should not be changing in this PR.

EricBolten (commented):
Left some comments on fixes.

steward/src/allocation.rs
```diff
@@ -31,9 +32,9 @@ pub struct TickWeight {
     pub weight: u32,
 }
 
-pub fn from_tick_weight(tick_weight: TickWeight) -> CellarTickInfo {
+pub fn from_tick_weight(tick_weight: TickWeight, cellars: CellarConfig) -> CellarTickInfo {
```
EricBolten: Nit: this is a singular `cellar` rather than `cellars`.

steward/src/allocation.rs
```diff
@@ -253,7 +254,10 @@ pub async fn direct_rebalance(
     let mut contract_state = UniswapV3CellarState::new(cellar_address, client);
     for tick_weight in tick_weight {
         if tick_weight.weight > 0 {
-            tick_info.push(from_tick_weight(tick_weight.clone()))
+            tick_info.push(from_tick_weight(
+                tick_weight.clone(),
+                config.cellars[0].clone(),
```
EricBolten: This technically works when there is only one cellar, but what we want is the cellar config whose cellar address matches the address from the request.

EricBolten (commented):
Some comments.

EricBolten (approved):
LGTM

PR #57, opened 2022-01-03
Created salt for allocation with Rand_core

hannydevelop:


steward/src/allocation.rs
```diff
@@ -148,6 +154,11 @@ pub fn to_allocation(
     pair_id: String,
     eth_gas_price: u64,
 ) -> somm::Allocation {
+    let salt = &generate_salt();
+    let salt = match std::str::from_utf8(salt) {
```
EricBolten: This implementation can possibly generate any valid UTF8 string, which is going to contain lots of characters that could be non-printable, hard to differentiate, etc.

Here is an example of taking a limited subset of characters that I proposed in an earlier PR:

https://github.com/PeggyJV/steward/pull/35#discussion_r770039091

steward/src/allocation.rs
```diff
@@ -52,7 +59,7 @@ pub async fn get_connections() -> Result<Connections, Error> {
     Ok(Connections { grpc, contact })
 }
 
-pub async fn decide_rebalance(tick_range: Vec<somm::TickRange>, cellar_address: H160) {
+pub async fn decide_rebalance(tick_range: Vec<somm::TickRange>, cellar_address: H160, pair_id: String) {
```
EricBolten: Let's rename `pair_id` to `cellar_id` here to reflect the different argument.

EricBolten (commented):
A few suggestions.

steward/src/allocation.rs
hannydevelop: So what @cbrit really wants is for me to totally remove pair_id from the decide_rebalance function and replace pair_id with cellar_address in line 31, 71, 105, 154

EricBolten: Yes, elsewhere when you are calling `decide_rebalance`, this new third argument is the `cellar_id` but it's still named `pair_id` here. Since we're replacing `pair_id` with `cellar_id` throughout the allocation code, let's rename the argument to `cellar_id` as well.

hannydevelop: fixed with this commit: https://github.com/PeggyJV/steward/pull/57/commits/944375d8703aeb9ab07c8c026461c33da3424dcf

steward/src/allocation.rs
hannydevelop: Fixed with this commit https://github.com/PeggyJV/steward/pull/57/commits/75849001e98d44b5c4309c707b52489196652085

steward/src/allocation.rs
```diff
@@ -145,17 +145,21 @@ pub async fn decide_rebalance(tick_range: Vec<somm::TickRange>, cellar_address:
...
             cellar: Some(somm::Cellar {
-                id: pair_id,
+                id: cellar_address,
                 tick_ranges: tick_ranges,
             }),
             current_price: eth_gas_price,
         }),
-        salt: "".to_string(), //TODO: Add salt,
+        salt: OsRng
+        .sample_iter(&Alphanumeric)
```
EricBolten: I believe this bit should be indented.

steward/src/allocation.rs
```diff
@@ -68,8 +69,7 @@ pub async fn decide_rebalance(tick_range: Vec<somm::TickRange>, cellar_address:
             }
         };
         let config = APP.config();
-        let pair_id = &config.cellars[0].pair_id;
-        let allocation = to_allocation(tick_range, pair_id.to_string(), eth_gas_price.as_u64());
+        let allocation = to_allocation(tick_range, cellar_address.to_string(), eth_gas_price.as_u64());
```
EricBolten: So there's a weird issue here with `H160`, `to_string` returns a truncated version of the output rather than the full hex string. This is how it was addressed in the gravity bridge:

https://github.com/PeggyJV/gravity-bridge/blob/4159c4fd56be1bf1ba586698cabeb5c8ca1f46c9/orchestrator/gravity_utils/src/ethereum.rs#L31-L33

EricBolten (commented):
Some suggested fixes.

steward/src/allocation.rs
EricBolten: Another alternative would be to just store the address as a string in `parse_cellar_id` and update `CellarId` to store a string. You could convert it to an `H160` elsewhere if it's needed. There's also a nicely named type in `ethers::types::Address` which is an `H160` in places where you want to use it to make it a bit more readable.

EricBolten (approved):
LGTM

PR #59, opened 2022-01-05
Handle parsing error for cellar_id

### LINKED ISSUE #38: Handling Data Provider Payload errors
Errors relating to request processing should be responded to with a `tonic::Status`. `tonic` has helpers for returning the appropriate status https://docs.rs/tonic/0.6.2/tonic/struct.Status.html

cbrit:
Closes #38 

Right now this was the only potential payload error that needed handling. Is there anything else that the DP should be informed of?

### DIFF
--- a/steward/src/cellars.rs
+++ b/steward/src/cellars.rs
@@ -1,4 +1,4 @@
-use crate::{error::Error, gas::CellarGas};
+use crate::{error::{Error, ErrorKind}, gas::CellarGas};
 use ethers::prelude::*;
 use std::{fmt, result::Result};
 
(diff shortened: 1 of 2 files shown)


EricBolten:
Some conflicts to resolve since merging the last PR unfortunately.

cbrit:
> Some conflicts to resolve since merging the last PR unfortunately.

fixed

steward/src/cellars.rs
```diff
@@ -23,12 +23,17 @@ pub async fn get_gas_price() -> Result<U256, Error> {
     CellarGas::etherscan_standard().await.map_err(|e| e.into())
 }
 
-fn parse_cellar_id(cellar_id: &str) -> CellarId {
-    let parts = cellar_id.split_at(cellar_id.chars().position(|c| c == ':').unwrap());
+fn parse_cellar_id(cellar_id: &str) -> Result<CellarId, String> {
+    let delimiter = match cellar_id.chars().position(|c| c == ':') {
```
EricBolten: I don't think you need to go through all of this, this should suffice:

https://doc.rust-lang.org/std/primitive.str.html#method.split

EricBolten (commented):
Added a suggestion.

steward/src/cellars.rs
cbrit: Improved and added unit tests

steward/src/cellars.rs
```diff
@@ -23,12 +23,48 @@ pub async fn get_gas_price() -> Result<U256, Error> {
     CellarGas::etherscan_standard().await.map_err(|e| e.into())
 }
 
-fn parse_cellar_id(cellar_id: &str) -> CellarId {
-    let parts = cellar_id.split_at(cellar_id.chars().position(|c| c == ':').unwrap());
-    let address: H160 = H160::from_slice(parts.1.as_bytes());
+fn parse_cellar_id(cellar_id: &str) -> Result<CellarId, String> {
```
EricBolten, line 26: Is there a specific error type we would prefer to be returning here rather than a String?

EricBolten (commented):
Comment around error typing.

steward/src/cellars.rs
cbrit: This was because the error message needs to be returned to the data provider via `tonic::Status::invalid_argument()` which expects a `String` error message. Any suggestions? Not sure we'd want to log this locally unless we are debugging.

EricBolten: It's probably fine in this case, but in general returning a basic type like String obfuscates the error condition and makes it more clunky to write code around handling it. Generally, a function should be acting like a black box -- if there is a more appropriate error type it can return, then a caller who needs a string representation of the error can format that at their discretion.

EricBolten (approved):
LGTM

PR #62, opened 2022-01-10
add data science client certs

### LINKED ISSUE #40: Certificate creation and cert/key storage

cbrit:
Closes #40


certs/client.crt
```diff
@@ -0,0 +1,13 @@
+-----BEGIN CERTIFICATE-----
```
EricBolten: We only need the CA cert to validate the connection. The data provider can update their cert or issue many for different machines as long as they are from the same CA.

EricBolten (commented):
Comment re: certs. It would also be wise to explain in the README or elsewhere why the config is set up the way it is to use this checked in client root CA.

steward/src/commands/cosmos_mode.rs
```diff
@@ -33,15 +33,19 @@ impl Runnable for CosmosSignerCmd {
...
                     std::process::exit(1)
                 });
 
             // run it
-            let addr = ([127, 0, 0, 1], STEWARD_PORT).into();
+            let port = match config.server.port {
+                Some(port) => port,
+                None => DEFAULT_STEWARD_PORT,
+            };
+            let addr = ([127, 0, 0, 1], port).into();
```
EricBolten: To answer a question I believe you asked out of band, yes, we should also make the listen address configurable.

EricBolten (commented):
One last comment re: config.

steward/src/commands/cosmos_mode.rs
cbrit: ~~Including domain name or just IP?~~

nevermind. Addressed this

EricBolten (approved):
LGTM

PR #63, opened 2022-01-13
removed pair_database and mongodb

hannydevelop:


cbrit (approved):

EricBolten:
We'll also need to delete https://github.com/PeggyJV/steward/blob/e5315df81b4ea5f2797d458a1bcf8ef28a101d27/steward/src/commands/print_config.rs#L21

if we can fully extract usage of Mongo from steward, I don't think anything is using `TimeRange` any more:
https://github.com/PeggyJV/steward/blob/main/steward/src/time_range.rs

and we can also then remove mongo from the Cargo.toml

PR #66, opened 2022-01-17
updated validators instructions

hannydevelop:
closes #58

PR #67, opened 2022-01-17
Collin/testing

### LINKED ISSUE #43: Test workflow and integration tests
Currently there is a basic (WIP) Dockerfile but no tests have been written nor a way to automatically run a container and execute tests.

cbrit:
Closes #43 

Originally branched off of bolten/testing-experiments so some commits are @EricBolten's

### Test instructions
 Prereqs: `make` and `docker` must be installed and the docker daemon running

1. run `make e2e_build_images`
2. once the test creates the steward containers you can track the debug logs in another terminal window by running `docker logs --follow steward0 | grep steward`

### New files:
- `steward.go` to match `orchestrator.go` and `validator.go`
- `steward_boostrap.sh`
- `tls` directory containing gRPC server/client tls certs and keys

### Additions to existing Go code:
https://github.com/PeggyJV/steward/blob/95df225516e8a4e7692521adc9d8f9a38ed9158d/integration_tests/chain.go#L182-L190

call steward init:
https://github.com/PeggyJV/steward/blob/95df225516e8a4e7692521adc9d8f9a38ed9158d/integration_tests/setup_test.go#L171

run steward:
https://github.com/PeggyJV/steward/blob/95df225516e8a4e7692521adc9d8f9a38ed9158d/integration_tests/setup_test.go#L708-L810

run rebalance:
https://github.com/PeggyJV/steward/blob/95df225516e8a4e7692521adc9d8f9a38ed9158d/integration_tests/rebalance_test.go#L66-L125


steward/src/somm_send.rs
```diff
@@ -86,16 +86,16 @@ pub async fn data_hash(
...
     let mut hasher = sha2::Sha256::new();
-    if let Some(cellar) = &allocation.clone().vote.unwrap().cellar {
+    if let Some(vote) = &allocation.clone().vote {
         let mut buf = BytesMut::new();
-        cellar.encode(&mut buf).unwrap();
-        let cellar_data = hex::encode(&buf).to_string();
-        let msg = format!("{}:{}:{}", allocation.salt, cellar_data, val_address);
+        vote.encode(&mut buf).unwrap();
+        let vote_data = hex::encode(&buf).to_string();
+        let msg = format!("{}:{}:{}", allocation.salt, vote_data, val_address);
```
cbrit, line 92: I think I found part of the cause of the hash mismatch problem but it still doesnt match. It looks like somm is encoding then hashing the entirety of the RebalanceVote whereas steward was just doing the Cellar part of the vote object. Could you confirm that this is a sane fix @EricBolten ? Compare with the following:

https://github.com/PeggyJV/sommelier/blob/4cc93e29f506ada4a572d845646ad3b3353a44dd/x/allocation/types/rebalance_vote.go#L28-L39

https://github.com/PeggyJV/sommelier/blob/4cc93e29f506ada4a572d845646ad3b3353a44dd/x/allocation/types/allocation.pb.go#L556-L564

EricBolten: Yes, it appears that the chain side is encoding the entire `RebalanceVote` and here we were only using the cellar field.

cbrit:
I noticed some logs don't render well with pretty print. I think it has to do with whether they are nested and have large JSON included. Changed a few `{:?}` to `{}` for debug logs

integration_tests/setup_test.go
```diff
@@ -0,0 +1,860 @@
...
+			Name:       steward.instanceName(),
+			NetworkID:  s.dockerNetwork.Network.ID,
+			Repository: "steward",
+			Tag:        "prebuilt",
+			Mounts: []string{
+				fmt.Sprintf("%s/:/root/steward", stewardCfgPath),
+			},
+			Env: []string{
+				"CELLAR_DRY_RUN=false",
+				"ETHERSCAN_API_KEY=W8C9NACKN284HFZCGNZA1CM4UHYZIMI4YN",
```
EricBolten: Probably don't want this checked in. You could extract whatever the current env's value is.

integration_tests/rebalance_test.go
```diff
@@ -0,0 +1,174 @@
...
+				s.T().Logf("sending request to %s", s.chain.validators[i].keyInfo.GetAddress())
+				cellarId := fmt.Sprintf("%s:%s", "ethereum", hardhatCellar.String())
+				request := RebalanceRequest{CellarId: cellarId, Data: data}
+				_, err = c.Rebalance(ctx, &request)
+				s.Require().NoError(err)
+			}
+
+			return true
+		}, 100*time.Second, 1*time.Second, "rebalance request took too long")
+
```
EricBolten, line 121: The tick ranges being eventually updated implies the success of the intermediate steps, but it would be good to add back test assertions to verify that the precommits and commits were successfully added to the chain. This is especially useful if one part of the overall flow breaks, otherwise it will be opaque to figure out what went wrong between data submission to steward and cellar update.

integration_tests/setup_test.go
```diff
@@ -0,0 +1,860 @@
...
+		runOpts := &dockertest.RunOptions{
+			Name:       steward.instanceName(),
+			NetworkID:  s.dockerNetwork.Network.ID,
+			Repository: "steward",
+			Tag:        "prebuilt",
+			Mounts: []string{
+				fmt.Sprintf("%s/:/root/steward", stewardCfgPath),
+			},
+			Env: []string{
+				"CELLAR_DRY_RUN=false",
```
EricBolten: You deleted the check for this environment variable in steward. Note that the readme still also references `CELLAR_DRY_RUN`.

steward/src/allocation.rs
```diff
@@ -50,114 +35,128 @@ pub async fn get_connections() -> Result<Connections, Error> {
...
+    let cosmos_delegate_key = config.load_deep_space_key(name.clone());
+    let cosmos_delegate_address = cosmos_delegate_key.to_address(&config.cosmos.prefix)?;
 
-        let cosmos_gas_price = config.cosmos.gas_price.as_tuple();
-        let fee = Coin {
-            denom: cosmos_gas_price.1,
-            amount: 0u32.into(),
-        };
+    debug!("querying the validator address based on the associated delegate address {}", cosmos_delegate_address);
+    let mut gravity_client = GravityQueryClient::connect(config.cosmos.grpc.clone()).await?;
```
EricBolten: I don't think this needs to be mutable. Would it make sense to add it to `get_connections()`?

steward/src/allocation.rs
```diff
@@ -50,114 +35,128 @@ pub async fn get_connections() -> Result<Connections, Error> {
...
+        }
+    }
+
+    match timeout(Duration::from_secs(100), async {
+        debug!("querying precommit");
+        loop {
+            let res = somm_send::query_allocation_precommits(grpc_client).await;
+            match res {
+                Ok(val) => {
+                    if val.len() > 0 && val.contains(allocation_precommit) {
```
EricBolten: You should only really need the `contains` check here, it should return very quickly if the length is zero.

EricBolten (commented):
Some comments. There's also a failing unit test which is being noticed by CI.

integration_tests/setup_test.go
cbrit: i knew I missed something in cleanup ty ser. Also CELLAR_DRY_RUN does nothing now

integration_tests/rebalance_test.go
cbrit: Are debug logs in the steward containers enough? I tried for a little while to keep those tests in but they were having weird timing issues. I can add them back though.

integration_tests/setup_test.go
cbrit: There's an open PR for the README that @hannydevelop is working on : https://github.com/PeggyJV/steward/pull/66

will remove the dry run reference

steward/src/allocation.rs
cbrit: I recall adding it because of a confusing panic. I'll try it again.

steward/src/allocation.rs
cbrit: It does; something about the `delegate_keys_by_orchestrator()` method called in the following method in `utils` requires it.

I can add it to connections

integration_tests/rebalance_test.go
EricBolten: It's helpful to have tests fail in specific ways to indicate which bit is incorrect, and also to fail quicker in CI if they are going to break. Logs aren't very helpful for that bit.

integration_tests/setup_test.go
cbrit: Pushed a fix

steward/src/allocation.rs
cbrit: This call borrows `client` as mutable: https://github.com/PeggyJV/steward/blob/7533448a1505dbfb8c219b7f884c2e08cf257eb8/steward/src/utils.rs#L30
signature
https://github.com/PeggyJV/gravity-bridge/blob/19ba1cc74d0fdfd7ba9f3cfcdad354c7259030bc/orchestrator/gravity_proto/src/prost/gravity.v1.rs#L1263-L1266

integration_tests/rebalance_test.go
cbrit: Done!

Had to remove the explicit precommit hash comparison test because it isn't possible for the test to know what salt steward used.

steward/src/allocation.rs
cbrit: fixed!

integration_tests/rebalance_test.go
```diff
@@ -124,6 +125,111 @@ func (s *IntegrationTestSuite) TestRebalance() {
 			return true
 		}, 100*time.Second, 1*time.Second, "rebalance request took too long")
 
+		s.T().Logf("wait for new vote period start")
```
EricBolten, line 122: Wait, does this actually work? If you are waiting until the next vote period after steward has submitted the precommits/commits they should be gone already.

integration_tests/rebalance_test.go
```diff
@@ -124,6 +125,111 @@ func (s *IntegrationTestSuite) TestRebalance() {
...
+
+				queryClient := types.NewQueryClient(clientCtx)
+				res, err := queryClient.QueryAllocationCommit(context.Background(), &types.QueryAllocationCommitRequest{Validator: sdk.ValAddress(val.keyInfo.GetAddress()).String(), Cellar: hardhatCellar.String()})
+				if err != nil {
+					return false
+				}
+				if res == nil {
+					return false
+				}
+				s.Require().Equal(res.Commit.Vote.Cellar.Id, commit.Vote.Cellar.Id, "cellar ids unequal")
```
EricBolten: The salt is revealed in the commit, you can verify that the contents match the precommit from earlier.

EricBolten (commented):
A few more comments.

steward/src/somm_send.rs
cbrit: It is technically a race but it works in my experience because the next vote period is ~15-20 seconds away when the rebalance request gets sent (at least on my machine). I feel like that is enough room to call it safe.. you?

integration_tests/rebalance_test.go
cbrit: It is technically a race but it works in my experience because the next vote period is ~15-20 seconds away when the rebalance request gets sent (at least on my machine). I feel like that is enough room to call it safe.. you? Can you run it and observe the steward debug logs to make sure you see the same?

EricBolten: I think this is only actually working because everything is submitted fast enough that it's in the same block, which will still be a vote period start. I would remove that delay bit, steward waits for the vote period start inside `decide_rebalance`.

Makefile
```diff
@@ -0,0 +1,54 @@
+.DEFAULT_GOAL := e2e_rebalance
+
+VALIDATOR_IMAGE := "ghcr.io/peggyjv/sommelier-sommelier:v3.0.0-pre2"
```
EricBolten: Both `VALIDATOR_IMAGE` and `ORCHESTRATOR_IMAGE` are pretty out of date now, let's give them a bump.

go.mod
```diff
@@ -0,0 +1,22 @@
...
+
+go 1.15
+
+require (
+	github.com/cosmos/cosmos-sdk v0.44.5
+	github.com/cosmos/go-bip39 v1.0.0
+	github.com/ethereum/go-ethereum v1.10.11
+	github.com/miguelmota/go-ethereum-hdwallet v0.1.1
+	github.com/ory/dockertest/v3 v3.7.0
+	github.com/peggyjv/gravity-bridge/module v0.3.0
```
EricBolten: The `peggyjv` deps here could use bumps as well.

integration_tests/ethereum/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+COPY package.json package.json
+COPY yarn.lock yarn.lock
+
+RUN yarn install --production=false
+RUN npm config set user 0
+
+COPY . .
+
+ENV ARCHIVE_NODE_URL=""
```
EricBolten, line 14: This line made me consider: should we create a README.md explaining the necessary dependencies and how to run the test? Feels like tribal knowledge right now.

integration_tests/rebalance_test.go
```diff
@@ -0,0 +1,285 @@
...
+			block, _ := pem.Decode(r)
+			rootCert, err := x509.ParseCertificate(block.Bytes)
+			s.Require().NoError(err)
+
+			rootPool.AddCert(rootCert)
+
+			tlsConfig := &tls.Config{
+				Certificates: []tls.Certificate{clientCert},
+				RootCAs:      rootPool,
+				ServerName:   "",
```
EricBolten: Nit: does this only need to be included if we are overriding the hostname verification check? I think the test certs are using localhost. It's probably not a huge deal if it's working correctly, but curious.

integration_tests/rebalance_test.go
EricBolten: Ah, I am realizing now that of course steward blocking on waiting for the vote period to start doesn't have any effect on the test's execution flow. This is probably still necessary.

integration_tests/rebalance_test.go
```diff
@@ -0,0 +1,285 @@
...
+				return false
+			}
+
+			return true
+		}, 105*time.Second, 1*time.Second, "new vote period never seen")
+
+		s.T().Logf("checking pre-commits for validators")
+		precommits := make([]types.AllocationPrecommit, 4)
+		for _, val := range s.chain.validators {
+			val := val
```
EricBolten: I don't think this is necessary.

integration_tests/rebalance_test.go
```diff
@@ -0,0 +1,285 @@
...
+					Validator: operator.String(),
+					Cellar:    hardhatCellar.String(),
+				})
+				if err != nil {
+					return false
+				}
+				if res == nil {
+					return false
+				}
+				s.Require().NoError(err, "unable to create precommit")
```
EricBolten: Nit: the error here would be from retrieving rather than creating a precommit.

integration_tests/rebalance_test.go
```diff
@@ -0,0 +1,285 @@
...
+			},
+				30*time.Second,
+				2*time.Second,
+				"pre-commit not found for validator %s",
+				val.keyInfo.GetAddress().String())
+		}
+
+		s.T().Logf("checking commits for validators")
+		for _, val := range s.chain.validators {
+			val := val
```
EricBolten: Ditto here, should be unnecessary.

integration_tests/rebalance_test.go
```diff
@@ -0,0 +1,285 @@
...
+			address := val.keyInfo.GetAddress()
+			operator := sdk.ValAddress(address)
+			s.Require().Eventuallyf(func() bool {
+				kb, err := val.keyring()
+				s.Require().NoError(err)
+				clientCtx, err := s.chain.clientContext("tcp://localhost:26657", &kb, "val", address)
+				s.Require().NoError(err)
+
+				queryClient := types.NewQueryClient(clientCtx)
```
EricBolten: Nit: not a blocker, but it would make these tests more readable to move the query client creation preamble to a function.

integration_tests/setup_test.go
```diff
@@ -0,0 +1,859 @@
...
+	for i, steward := range s.chain.stewards {
+		stewardCfg := fmt.Sprintf(`[cosmos]
+key_derivation_path = "m/44'/118'/1'/0/0"
+grpc = "http://%s:9090"
+prefix = "somm"
+msg_batch_size = 5
+gas_adjustment = 1.0
+
+[cosmos.gas_price]
+amount = 1000000000
```
EricBolten, line 720: Nit: I guess if it works it works but this seems really high!

steward/src/allocation.rs
```diff
@@ -41,124 +30,139 @@ pub async fn allocation_precommit(
...
+    debug!("getting eth gas price");
+    let eth_gas_price = match cellars::get_gas_price().await {
+        Ok(gp) => {
+            debug!("eth gas price is {}", gp);
+            gp
+        }
+        Err(err) => {
+            error!("failed to get cellar gas price: {}", err);
+            // TO-DO handle this better
+            0.into()
```
EricBolten, line 55: Nit: I remember you mentioning you wanted to update this but it can wait for another PR.

EricBolten (commented):
Added some comments, mostly tiny stuff and nits. Would be good to verify things still work as expected against the latest releases.

integration_tests/setup_test.go
cbrit: copied from somewhere ¯\_(ツ)_/¯

steward/src/allocation.rs
cbrit: working on it in another branch!

integration_tests/setup_test.go
cbrit: Actually it fails with an insufficient fees error with less. I didn't investigate to see why that's the required fee.

integration_tests/rebalance_test.go
cbrit: It also would never get reached because of the if statement above line 169

integration_tests/rebalance_test.go
cbrit: remove and it works

integration_tests/ethereum/Dockerfile
cbrit: I could leave a comment? Since the `integration_tests` dir is replicated across three repos right now maybe we should move it to a single test repo and document it there?

EricBolten (approved):
Remaining outstanding items are nits or minor. The version bump on the sommelier dependency requires external fixes and shouldn't be a blocker. This LGTM.

EricBolten:
Oh, whoops, just noticed there are conflicts. Those gotta be resolved before we merge.

PR #73, opened 2022-01-28
Collin/eth gas fallback

cbrit:
If the `ETHERSCAN_API_KEY` env var is missing, or the request to Etherscan fails, steward will estimate gas via `eth_gasPrice` using ethers.


steward/src/utils.rs
```diff
@@ -32,3 +38,111 @@ pub async fn get_delegates_keys_by_orchestrator(
...
+fn check_scheme(input: &Url, original_string: &str) {
+    if !(input.scheme() == "http" || input.scheme() == "https") {
+        panic!(
+            "Your url {} has an invalid scheme, please chose http or https",
+            original_string
+        )
+    }
+}
+
+pub async fn get_eth_provider(eth_rpc_url: &str) -> Result<Provider<Http>, Error> {
```
EricBolten, line 62: I don't think we need to copy-paste this sort of mess of a setup from the bridge code. We should simply try or fail given the RPC URL we are provided.

EricBolten (commented):
Some thoughts on instantiating the Ethereum provider.

steward/src/utils.rs
cbrit: I'm glad you said that lol.

steward/src/gas.rs
```diff
@@ -52,4 +59,19 @@ impl CellarGas {
         let data = etherchain_oracle.fetch().await;
         data
     }
+
+    pub fn pad(price: U256) -> Result<U256, Error> {
```
EricBolten: Nit: I don't think the term "pad" is accurate for multiplying by a floating point.

steward/src/gas.rs
```diff
@@ -52,4 +59,19 @@ impl CellarGas {
...
+            Some(p) => p,
+            None => {
+                return Err(ErrorKind::GasOracle
+                    .context("failed to downcast gas price estimate")
+                    .into())
+            }
+        };
+        let price = price * config.ethereum.gas_price_multiplier;
+
+        Ok((price as u64).into())
```
EricBolten: I believe you'd want to cast this as a `u128` rather than `u64` in terms of their relative max values.

steward/src/utils.rs
```diff
@@ -32,3 +37,27 @@ pub async fn get_delegates_keys_by_orchestrator(
 
     Ok(keys)
 }
+
+fn check_scheme(input: &Url, original_string: &str) -> Result<(), Error> {
```
EricBolten: Is this not something that would be captured in an error from `Provider::<Http>::try_from`?

steward/src/utils.rs
```diff
@@ -32,3 +37,27 @@ pub async fn get_delegates_keys_by_orchestrator(
...
+    if input.scheme() != "http" && input.scheme() != "https" {
+        return Err(ErrorKind::Config
+            .context(format!(
+                "Your url {} has an invalid scheme, please chose http or https",
```
EricBolten: Nit: `chose` -> `choose`.

steward/src/cellars.rs
```diff
@@ -17,7 +18,19 @@ impl std::fmt::Display for CellarId {
 }
 
 pub async fn get_gas_price() -> Result<U256, Error> {
-    CellarGas::etherscan_standard().await.map_err(|e| e.into())
+    if std::env::var("ETHERSCAN_API_KEY").is_ok() {
+        match CellarGas::etherscan_standard().await {
```
EricBolten, line 22: If we have any desire to run against anything other than Ethereum mainnet, we might also want to port the chain lookup code from the bridge as well:

https://github.com/PeggyJV/gravity-bridge/blob/fef4573e412be03a622496b0bd24365dd95265b8/orchestrator/ethereum_gravity/src/utils.rs#L141-L177

EricBolten (commented):
Some comments. It seems after merging main you have some build issues as well.

steward/src/error.rs
```diff
@@ -2,7 +2,8 @@
 
 use abscissa_core::error::{BoxError, Context};
 use deep_space::error::{AddressError, CosmosGrpcError, PrivateKeyError};
-use ethers::{contract::ContractError, middleware::gas_oracle::GasOracleError, prelude::*};
+use ethers::prelude::*;
+use ethers::prelude::{errors::EtherscanError, gas_oracle::GasOracleError};
```
EricBolten, line 6: Nit: a star import and specific imports shouldn't both be necessary.

steward/src/gas.rs
```diff
@@ -52,4 +76,13 @@ impl CellarGas {
         let data = etherchain_oracle.fetch().await;
         data
     }
+
+    async fn get_etherscan_client() -> Result<Client, Error> {
+        let config = APP.config();
+        let url = &config.ethereum.rpc;
+        let provider = crate::utils::get_eth_provider(url.as_str()).await?;
+        let chain = utils::get_chain(provider.clone()).await?;
```
EricBolten, line 84: Nit: I think it's a bit more confusing for this and the above line to have fully specified crate paths rather than just importing the functions themselves, especially with two different "utils" packages.

steward/src/cellars.rs
```diff
@@ -17,7 +18,20 @@ impl std::fmt::Display for CellarId {
...
+            Ok(gas) => return Ok(gas),
+            Err(err) => {
+                warn!("failed to retrieve gas estimate from etherscan: {}", err);
+            }
+        }
+    }
+
+    let config = APP.config();
+    let url = &config.ethereum.rpc;
+    let provider = crate::utils::get_eth_provider(url.as_str()).await?;
```
EricBolten, line 32: Nit: we repeat this three line setup in a few places, probably calls for a function to get the provider.

EricBolten (approved):
Only comments are nits, this LGTM.

PR #75, opened 2022-02-07
Improving UX for Steward

hannydevelop:


steward/src/commands/keys/cosmos/add.rs
```diff
@@ -6,9 +6,13 @@ use rand_core::OsRng;
 use signatory::FsKeyStore;
 use std::path;
 
+/// Gorc keys cosmos add [name]
 #[derive(Command, Debug, Default, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Create a new Cosmos Key.\n This command creates a new Cosmos key with a name as a String.\n It has an overwrite option, a boolean which can be true or false."
+)]
 pub struct AddCosmosKeyCmd {
-    pub args: Vec<String>,
```
philipjames44, line 11: I think args might be the more apt identifier if we expect to have to process any other arguments, otherwise if we're just planning on processing the name perhaps this is better as a stand alone string instead of a vector?

steward/src/commands/keys/cosmos/add.rs
```diff
@@ -6,9 +6,13 @@ use rand_core::OsRng;
...
+    pub name: Vec<String>,
```
cbrit: Can this command take more than one name or should it just be a `String`?
EDIT: It looks like we are just taking the first entry (line 29) so it shouldn't be necessary for it to be a vec I think

steward/src/commands/keys/cosmos/delete.rs
```diff
@@ -3,20 +3,24 @@ use abscissa_core::{clap::Parser, Application, Command, Runnable};
...
 use std::path::Path;
 
+/// Gorc keys cosmos delete [name]
 #[derive(Command, Debug, Default, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Delete a Cosmos Key.\n This command deletes a Cosmos key from your keystore it takes a name as a String."
+)]
 pub struct DeleteCosmosKeyCmd {
-    pub args: Vec<String>,
+    pub name: Vec<String>,
```
cbrit: Same question as with the add command

steward/src/commands/cosmos_to_eth.rs
```diff
@@ -10,12 +10,13 @@ use std::{process::exit, time::Duration};
 
 const TIMEOUT: Duration = Duration::from_secs(60);
 
-/// Send Cosmos to the Eth chain
+/// Send Cosmos to the Eth chain. args is a Vector of Strings and flag_no_batch is boolean
```
EricBolten: We shouldn't need to include this typing information in the command-level documentation when the flag details should explain it below.

steward/src/commands/cosmos_to_eth.rs
```diff
@@ -10,12 +10,13 @@ use std::{process::exit, time::Duration};
...
 #[derive(Command, Debug, Default, Parser)]
 pub struct CosmosToEthCmd {
+    /// <gravity_denom> <amount> <cosmos_key> <eth_dest> <times>
     #[clap(short, long)]
     pub args: Vec<String>,
-
+    /// boolean
```
EricBolten: Knowing the expected type is good, but it should also explain what it means to set `flag_no_batch`.

steward/src/commands/deploy/erc20.rs
```diff
@@ -13,12 +13,17 @@ use tokio::time::sleep as delay_for;
 
 /// Deploy Erc20
 #[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Deploy ERC20 tokens to the Sommelier Chain via the Gravity Bridge.\n This command takes a Vector of Strings called denom as args.\n Options are ethereum-key and gas-multiplier which is a 64-bit floating point type, f64."
```
EricBolten: Same deal here and elsewhere, we don't need to list flags/types which will be printed out automatically based on the struct.

steward/src/commands/keys/eth/delete.rs
```diff
@@ -3,21 +3,22 @@ use abscissa_core::{clap::Parser, Application, Command, Runnable};
 use signatory::FsKeyStore;
 use std::path;
 
-/// Delete Key
+/// Gorc keys eth delete [name]
```
EricBolten: We should avoid referencing `gorc` inside steward since it's a separate CLI implementation. I also don't think this type of comment is necessary as running `--help` for a specific command implies the format already (I believe this is autofilled in the "USAGE" section).

steward/src/commands/tx/cosmos.rs
```diff
@@ -25,9 +25,13 @@ impl Runnable for Cosmos {
...
 }
 
+/// Send tx from Cosmos to Ethereum chain
 #[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Send transactions from Cosmos to Ethereum chain.\n This command sends tx from the Cosmos chain to the Eth chain.\n It takes a free vector of strings as required flags for Cosmos key, Eth key and amount."
+)]
 pub struct SendToEth {
-    free: Vec<String>,
+    required_flags: Vec<String>,
```
EricBolten: If we are requiring flags, rather than using a vector of strings, we should create elements inside this struct for each flag with no default value.

EricBolten (commented):
Some comments.

steward/src/commands/keys/cosmos/add.rs
hannydevelop: Hey Phil, I think it'll be best to use individually process the arguments instead of using a vector. This because using a vector will render this:

```
ARGS:
    <ARGS>...  
```
However, for commands with multiple args like this:
https://github.com/PeggyJV/steward/blob/272a5a294af62b5a2398d9bec656938482a96bee/steward/src/commands/cosmos_to_eth.rs#L46-L53

The user won't be able to know what arguments are needed since it'll still display:

```
ARGS:
    <ARGS>...  
```

philipjames44: Agreed, individually processing the arguments sounds sensible.

steward/src/commands/keys/cosmos/add.rs
hannydevelop: I think Phil and Eric both said something about this,  we'll be removing all vecs for all options and replace them with individual arguments. Also, FYI this PR will distrupt Steward, especially the test.

steward/src/commands/allow_erc20.rs
```diff
@@ -10,11 +10,19 @@ use crate::{erc20::Erc20State, gas::CellarGas, prelude::*};
 
 /// Allow Erc20 Token to interact with cellar contract
 #[derive(Command, Debug, Default, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Allow ERC20 tokens to interact with Cellar contract.\n This command approves an Eth transaction with a Cellar address. \n It takes the Cellar address, Eth address and transaction amount."
+)]
 pub struct AllowERC20 {
+    /// H160 Cellar address you want to interact with.
     #[clap(short = 'C', long)]
     cellar_address: H160,
```
EricBolten, line 19: I see this is the way it already was, but is `clap` doing something here to convert the string into an H160 automatically?

steward/src/commands/allow_erc20.rs
```diff
@@ -10,11 +10,19 @@ use crate::{erc20::Erc20State, gas::CellarGas, prelude::*};
 
 /// Allow Erc20 Token to interact with cellar contract
 #[derive(Command, Debug, Default, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Allow ERC20 tokens to interact with Cellar contract.\n This command approves an Eth transaction with a Cellar address. \n It takes the Cellar address, Eth address and transaction amount."
+)]
 pub struct AllowERC20 {
+    /// H160 Cellar address you want to interact with.
```
EricBolten: We can drop the `H160` bit in the description, addresses all have the same format and that's an internal type.

steward/src/commands/cosmos_mode.rs
```diff
@@ -10,7 +10,11 @@ use abscissa_core::{clap::Parser, config, Command, FrameworkError, Runnable};
 use std::result::Result;
 use steward_proto::uniswapv3::server::UniswapV3CellarAllocatorServer;
 
+/// Cosmos Signer, start allocation module
 #[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Cosmos Signer, start allocation module.\n This command starts Steward using the allocation module."
```
EricBolten: I think a more generic description might be appropriate here, something along the lines of "Run Steward as a server that will send updates to the Sommelier chain" or similar.

steward/src/commands/cosmos_to_eth.rs
```diff
@@ -10,12 +10,33 @@ use std::{process::exit, time::Duration};
...
+
+    /// H160 Ethereum address
+    #[clap(short, long)]
+    eth_dest: String,
+
+    /// The number of times transactions should repeat itself, default is 1.
+    #[clap(short, long, default_value = "1")]
+    times: String,
+
+    /// Boolean, True if you want to wait until someone requests a batch for this token type and False if you want to request a batch to push transaction along immediately.
```
EricBolten, line 39: We might want to consider changing how this flag works, since a double negative is kind of confusing, i.e. `flag_no_batch` being `false` means send a batch. Perhaps changing this to `request_batch` and having `true` send the request would be easier to understand.

steward/src/commands/eth_to_cosmos.rs
```diff
@@ -14,24 +14,49 @@ const TIMEOUT: Duration = Duration::from_secs(60);
...
+
+    /// Tx amount.
+    #[clap(short, long)]
+    init_amount: String,
+
+    /// Eth keyname, takes a String.
+    #[clap(short, long)]
+    ethereum_key: String,
+
+    /// H160 Cosmos address
```
EricBolten: Aside from the general comment around not needing to specify `H160`, Cosmos addresses will not be in that format anyway.

steward/src/commands/keys/cosmos/add.rs
```diff
@@ -6,24 +6,28 @@ use rand_core::OsRng;
 use signatory::FsKeyStore;
 use std::path;
 
+/// Steward keys cosmos add [name]
```
EricBolten: The help message will specify the format of the command. These descriptors are more for a documentation-level description of what this command is doing.

steward/src/commands/keys/cosmos/add.rs
```diff
@@ -6,24 +6,28 @@ use rand_core::OsRng;
...
 #[derive(Command, Debug, Default, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Create a new Cosmos Key.\n This command creates a new Cosmos key. It has an overwrite option, which if set to true, overwrites\n an existing key in the keystore with similar keyname."
```
EricBolten: Saying the same thing twice in the first two lines here.

steward/src/commands/keys/cosmos/delete.rs
```diff
@@ -3,22 +3,26 @@ use abscissa_core::{clap::Parser, Application, Command, Runnable};
...
+    long_about = "DESCRIPTION \n\n Delete a Cosmos Key.\n This command deletes a Cosmos key from your keystore when provided with the keyname."
+)]
 pub struct DeleteCosmosKeyCmd {
-    pub args: Vec<String>,
+    /// Cosmos keyname in keystore, takes a String
+    pub name: String,
 }
 
-/// The `gork keys cosmos delete [name] ` subcommand: delete the given key
+/// The `steward keys cosmos delete [name] ` subcommand: delete the given key
```
EricBolten, line 15: Similarly this bit is probably not necessary for documentation.

steward/src/commands/keys/cosmos/delete.rs
```diff
@@ -3,22 +3,26 @@ use abscissa_core::{clap::Parser, Application, Command, Runnable};
...
     fn run(&self) {
         let config = APP.config();
         // Path where key is stored.
         let keystore = Path::new(&config.keystore);
         let keystore = signatory::FsKeyStore::create_or_open(keystore).unwrap();
-        // Collect key name from args.
-        let name = self.args.get(0).expect("name is required");
-        let name = name.parse().expect("Could not parse name");
-        // Delete keyname after locating file from path and key name.
+        // Collect keyname from name arguement.
```
EricBolten: We can drop the comments in this `run` method, the code is straightforward enough.

steward/src/commands/keys/cosmos/recover.rs
```diff
@@ -5,34 +5,42 @@ use k256::pkcs8::ToPrivateKey;
...
 #[derive(Command, Debug, Default, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Recover an external Cosmos key.\n This command will recover a Cosmos key, storing it in the keystore. \n It takes a keyname and bip39-mnemonic."
+)]
 pub struct RecoverCosmosKeyCmd {
-    pub args: Vec<String>,
+    /// Cosmos keyname, takes a String
+    pub name: String,
 
+    /// Overwrite key with similar name in the keystore when set to true. Takes a Boolean.
```
EricBolten: I think here by "similar" we mean the key with precisely that name.

steward/src/commands/keys/cosmos/show.rs
```diff
@@ -1,16 +1,20 @@
...
 use abscissa_core::{clap::Parser, Application, Command, Runnable};
 
+/// Gorc keys cosmos show [name]
 #[derive(Command, Debug, Default, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Show details of a Cosmos key in the keystore.\n This command shows details of a Cosmos key in the keystore, it takes the name of the key."
+)]
 pub struct ShowCosmosKeyCmd {
-    pub args: Vec<String>,
+    pub name: String,
```
EricBolten, line 10: Should we add a documentation comment for this field?

steward/src/commands/query/cosmos.rs
```diff
@@ -17,33 +17,41 @@ impl Runnable for Cosmos {
...
 }
 
+/// Query Cosmos balance
 #[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Query Cosmos balance.\n This command queries the Cosmos balance, taking the name of the key as a String."
+)]
 pub struct Balance {
-    free: Vec<String>,
+    key_name: Vec<String>,
```
EricBolten: Still using a vector here rather than a plain `String` (and elsewhere in this file).

steward/src/commands/query/eth.rs
```diff
@@ -13,18 +13,22 @@ pub enum Eth {
...
 }
 
+/// Query Eth balance
 #[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Query Eth balance.\n This command queries the Eth balance, taking the name of the key as a String."
+)]
 pub struct Balance {
-    free: Vec<String>,
+    key_name: Vec<String>,
```
EricBolten: Also still using a vector here.

steward/src/commands/set_validator.rs
```diff
@@ -6,12 +6,14 @@ use signatory::FsKeyStore;
...
 
-/// `set-validator` subcommand
+/// Set-validator subcommand
 #[derive(Command, Debug, Parser)]
 pub struct SetValidatorCmd {
+    /// H160 Validator Address
     #[clap(short = 'v', long)]
     validator: H160,
 
+    ///
```
EricBolten: Should there be something here?

steward/src/commands/tx/cosmos.rs
EricBolten: Following up on this comment as you addressed the "required args" format in most of the other commands but not this one. Alternatively, if these are duplicate commands that are unfinished, perhaps we should simply delete the `tx` subcommand structure entirely.

steward/src/commands/tx/cosmos.rs
```diff
@@ -148,9 +152,13 @@ impl Runnable for SendToEth {
...
 }
 
+/// Send tx from across the Cosmos chain
 #[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Send transactions from across the Cosmos chain.\n This command sends tx from across the Cosmos chain.\n It takes a free vector of strings as required flags for sender's key, reciever's key and amount."
+)]
 pub struct Send {
-    free: Vec<String>,
+    required_flags: Vec<String>,
```
EricBolten: Same here re: "required flags".

EricBolten (commented):
Some comments.

steward/src/commands/allow_erc20.rs
hannydevelop: Yes, I think the `syn` crate does that. This is the [parser mod](https://github.com/clap-rs/clap/blob/master/clap_derive/src/derives/parser.rs) https://crates.io/crates/syn

steward/src/commands/allow_erc20.rs
hannydevelop: done!

steward/src/commands/cosmos_mode.rs
hannydevelop: done!

steward/src/commands/eth_to_cosmos.rs
hannydevelop: What format should the Cosmos address be in? I removed H160 from the description

steward/src/commands/keys/cosmos/add.rs
hannydevelop: Done, removed this in all of keys command

steward/src/commands/keys/cosmos/add.rs
hannydevelop: Edited the command description

steward/src/commands/keys/cosmos/delete.rs
hannydevelop: done!

steward/src/commands/keys/cosmos/delete.rs
hannydevelop: Done!

steward/src/commands/keys/cosmos/recover.rs
hannydevelop: yes, is there any way you'd like me to improve the description?

steward/src/commands/keys/cosmos/show.rs
hannydevelop: done!

steward/src/commands/set_validator.rs
hannydevelop: yes, edited it with a description

steward/src/commands/cosmos_to_eth.rs
hannydevelop: please can you checkout if what I did solves the problem

EricBolten: Nit: this is different but also works, as the true condition makes more sense. The only thing I'd suggest is having it say `wait_for_batch`.

steward/src/commands/cosmos_to_eth.rs
```diff
@@ -10,14 +10,35 @@ use std::{process::exit, time::Duration};
...
 const TIMEOUT: Duration = Duration::from_secs(60);
 
-/// Send Cosmos to the Eth chain
+/// Send Cosmos to the Eth chain.
 #[derive(Command, Debug, Default, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Send Cosmos token to Eth chain.\n This command sends Cosmos token to the Eth chain via the Gravity bridge. \n This command takes the Gravity denom, tx amount, Cosmos keyname, Eth destination, number of times \n transaction should be made and if the transaction should be made immediately or wait for the next \n batch."
+)]
 pub struct CosmosToEthCmd {
+    /// Gravity denom, takes a String.
```
EricBolten: It's not necessary to specify `String` types here, everything coming from the command line is sent as a `String` from the user's perspective, it's only worth mentioning what the type is if the format is relevant (e.g. addresses, booleans, etc.).

steward/src/commands/eth_to_cosmos.rs
```diff
@@ -14,24 +14,49 @@ const TIMEOUT: Duration = Duration::from_secs(60);
 
 /// Send Ethereum to Cosmos
 #[derive(Command, Debug, Default, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Send Eth token to Cosmos chain.\n This command sends Eth token to the Cosmos chain via the Gravity bridge. \n It takes the tx amount, Eth keyname, contract address, Cosmos token destination, number of times \n transaction should be made and if the transaction should be made immediately or wait for the next \n batch."
```
EricBolten: I think the was partially copy-pasted from the other command. It should include `erc20_address` and there is no bool here for waiting for a batch, transfers from Ethereum to Cosmos are unbatched.

steward/src/commands/keys/cosmos/add.rs
```diff
@@ -7,23 +7,26 @@ use signatory::FsKeyStore;
 use std::path;
 
 #[derive(Command, Debug, Default, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Create a new Cosmos Key.\n This command creates a new Cosmos key when provided an overwrite option, which if set to true, overwrites\n an existing key in the keystore with similar keyname."
```
EricBolten: Same comment here re: other usages of "similar". What we mean to say is that it overwrites an existing key in the keystore with the same name.

steward/src/commands/keys/cosmos/recover.rs
EricBolten: Sure, it's the same name, not a similar name. Perhaps "Overwrites the specified key in the keystore" or something along those lines.

steward/src/commands/keys/eth/add.rs
```diff
@@ -6,25 +6,27 @@ use rand_core::OsRng;
 use signatory::FsKeyStore;
 use std::path;
 
-/// Create a new Key
 #[derive(Command, Debug, Default, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Create a new Eth Key.\n This command creates a new Eth key. It has an overwrite option, which if set to true, overwrites\n an existing key in the keystore with similar keyname."
```
EricBolten: Ditto here with the "similar" comment, just pointing out all the cases.

steward/src/commands/set_validator.rs
```diff
@@ -6,12 +6,14 @@ use signatory::FsKeyStore;
 
 use crate::{cellars::uniswapv3::UniswapV3CellarState, gas::CellarGas, prelude::*};
 
-/// `set-validator` subcommand
+/// Set-validator subcommand
 #[derive(Command, Debug, Parser)]
 pub struct SetValidatorCmd {
+    /// H160 Validator Address
```
EricBolten: Should drop the "H160" description in the comment here as well (not meaningful from the user's perspective, their input is always a String). If it converts the type on the backend it's fine, but the user doesn't need to know that.

steward/src/commands/sign_delegate_keys.rs
```diff
@@ -6,21 +6,28 @@ use std::time::Duration;
 /// Sign delegate keys
 #[derive(Command, Debug, Default, Parser)]
 pub struct SignDelegateKeysCmd {
-    pub args: Vec<String>,
+    /// Ethereum keyname
+    #[clap(short, long)]
+    ethereum_key: String,
+
+    /// H160 Validator address
```
EricBolten: Ditto here on H160.

steward/src/commands/tx/cosmos.rs
```diff
@@ -25,9 +25,20 @@ impl Runnable for Cosmos {
     }
 }
 
+/// Send tx from Cosmos to Ethereum chain
 #[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Send transactions from Cosmos to Ethereum chain.\n This command sends tx from the Cosmos chain to the Eth chain.\n It takes a free vector of strings as required flags for Cosmos key, Eth key and amount."
```
EricBolten: Since we've converted to flags none of the `tx` commands take a free vector of strings any more.

EricBolten (commented):
Left some further comments. We're getting close to done! Thanks for continuing to refine this.

steward/src/commands/query/cosmos.rs
```diff
@@ -17,33 +17,43 @@ impl Runnable for Cosmos {
...
     #[clap(short, long)]
     help: bool,
 }
 
 impl Runnable for GravityKeys {
     /// Start the application.
     fn run(&self) {
-        assert!(self.free.len() == 1);
-        let _key_name = self.free[0].clone();
+        assert!(self.key_name.len() == 1);
```
EricBolten: This will break for any `key_name` that is longer than one character since this is no longer a vector.

steward/src/commands/query/eth.rs
```diff
@@ -13,18 +13,23 @@ pub enum Eth {
...
 
     #[clap(short, long)]
     help: bool,
 }
 
 impl Runnable for Balance {
     fn run(&self) {
-        assert!(self.free.len() == 1);
-        let _key_name = self.free[0].clone();
+        assert!(self.key_name.len() == 1);
```
EricBolten: Same thing here with the assertion on `key_name`.

steward/src/commands/tx/cosmos.rs
EricBolten: This bit still refers to the flags here as a "free vector of strings".

steward/src/commands/query/cosmos.rs
```diff
@@ -17,33 +17,43 @@ impl Runnable for Cosmos {
...
 
     #[clap(short, long)]
     help: bool,
 }
 
 impl Runnable for Balance {
     fn run(&self) {
-        assert!(self.free.len() == 1);
-        let _key_name = self.free[0].clone();
+        assert!(self.key_name.len() == 1);
```
EricBolten: Same here, need to delete this assertion.

EricBolten (commented):
Just a few more comments on remaining issues.

steward/src/commands/fund_cellar.rs
```diff
@@ -15,11 +15,14 @@ use steward_abi::cellar_uniswap::*;
 
 /// Command to fund Cellars
 #[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Fund Cellar.\n This command funds the required Cellar by loading the Cellar config. \n It takes the amounts for the Cellar pool pair."
+)]
 pub struct FundCellarCmd {
-    #[clap(short = 'i', long)]
-    pub cellar_id: u32,
```
EricBolten, line 20: I don't think we can delete this. The command is just retrieving the first element (`cellars.get(0)`) which means you can't really pick what cellar you're funding. I think the likely intent here is that we take `cellar_id` as a string and find the config that matches (just like we are doing when handling allocation messages).

EricBolten (commented):
Noticed one last thing, this should probably be the final update.

EricBolten (approved):
LGTM

PR #82, opened 2022-02-22
improve cosmos-signer error handling

### LINKED ISSUE #39: Internal Error Logging and Handling
Gracefully log and handle any errors that would currently panic because of `unwrap()`. This issue's primary focus is internal server/processing/networking errors and not things that the Data Provider necessarily needs to hear about in a response. Though if we fail to process a request because of a server issue, perhaps we should send them a 500?

cbrit:
Tried to improve error handling around some `unwrap()` calls that needed it; it seems like we lean toward using `expect()` when the failure is due to an invalid config value so I also continued that pattern. Happy to discuss.

### DIFF
--- a/steward/src/allocation.rs
+++ b/steward/src/allocation.rs
@@ -246,8 +246,8 @@ pub async fn direct_rebalance(
     let wallet: LocalWallet = Wallet::from(key);
 
     let eth_host = config.ethereum.rpc.clone();
-    let provider = Provider::<Http>::try_from(eth_host.clone()).unwrap();
-
+    let provider = Provider::<Http>::try_from(eth_host.clone())
+        .unwrap_or_else(|_| panic!("could not get provider from eth_host {}", eth_host));
     let chain_id = provider
         .get_chainid()
         .await
--- a/steward/src/somm_send.rs
+++ b/steward/src/somm_send.rs
@@ -92,21 +92,20 @@ pub async fn data_hash(
         let msg = format!("{}:{}:{}", allocation.salt, vote_data, val_address);
         hasher.update(msg.as_bytes());
 
+        if vote.cellar.is_none() {
+            return Err(ErrorKind::AllocationError
+                .context("vote has empty cellar field")
+                .into());
+        }
+        let cellar_id = vote.cellar.clone().unwrap().id;
+
         return Ok(AllocationPrecommit {
             hash: hasher.finalize().to_vec(),
-            cellar_id: vote
-                .cellar
-                .clone()
-                .ok_or_else::<Error, _>(|| {
-                    ErrorKind::AllocationError
-                        .context("vote has empty cellar field")
-                        .into()
-                })?
-                .id,
+            cellar_id,
         });
     }
     Err(ErrorKind::AllocationError
-        .context("No cellar".to_string())
+        .context("no vote".to_string())
         .into())
 }
 
--- a/steward/src/somm_send.rs
+++ b/steward/src/somm_send.rs
@@ -60,7 +60,9 @@ async fn __send_messages(
     fee: Coin,
     messages: Vec<Msg>,
 ) -> Result<TxResponse, CosmosGrpcError> {
-    let cosmos_address = cosmos_key.to_address(&contact.get_prefix()).unwrap();
+    let cosmos_address = cosmos_key
+        .to_address(&contact.get_prefix())
+        .expect("failed to get address from cosmos key");
 
     let fee = Fee {
         amount: vec![fee],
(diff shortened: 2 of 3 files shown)


steward/src/allocation.rs
```diff
@@ -246,8 +246,8 @@ pub async fn direct_rebalance(
     let wallet: LocalWallet = Wallet::from(key);
 
     let eth_host = config.ethereum.rpc.clone();
-    let provider = Provider::<Http>::try_from(eth_host.clone()).unwrap();
-
+    let provider = Provider::<Http>::try_from(eth_host.clone())
+        .unwrap_or_else(|_| panic!("could not get provider from eth_host {}", eth_host));
```
philipjames44, line 250: Why unwrap_or_else vs expect in this specific instance?

EricBolten: Yes, this is what `expect` is intended to do, `panic!` with a message if an unwrap fails. `unwrap_or_else` is for returning an alternative/default value.

steward/src/somm_send.rs
```diff
@@ -87,14 +89,18 @@ pub async fn data_hash(
...
         let msg = format!("{}:{}:{}", allocation.salt, vote_data, val_address);
         hasher.update(msg.as_bytes());
 
         return Ok(AllocationPrecommit {
             hash: hasher.finalize().to_vec(),
-            cellar_id: vote.cellar.clone().unwrap().id,
+            cellar_id: vote
+                .cellar
+                .clone()
+                .expect("vote contains no 'cellar' value")
```
EricBolten: In a function that returns a result, we should probably do that here as well rather than `panic!`.

steward/src/somm_send.rs
```diff
@@ -60,7 +60,9 @@ async fn __send_messages(
     fee: Coin,
     messages: Vec<Msg>,
 ) -> Result<TxResponse, CosmosGrpcError> {
-    let cosmos_address = cosmos_key.to_address(&contact.get_prefix()).unwrap();
+    let cosmos_address = cosmos_key
+        .to_address(&contact.get_prefix())
+        .expect("failed to get address from cosmos key");
```
EricBolten: Similarly here, we should probably bubble errors up. It would be preferable to validate configs at startup time rather than later on. For example, if we know we have some Cosmos key we'll be using all the time, we can generate the address for it at config processing time and store it so we aren't constantly re-running these operations and checks.

EricBolten (commented):
Some comments.

steward/src/allocation.rs
cbrit: This was an expect wrapping a `fmt` call because of the string interpolation, but clippy wanted me to do it this way rather than use expect(fmt!()). I could just make the error message less specific and use expect?

cbrit:
Currently blocked because a dependency of our `deep_space` fork called `anomaly` was yanked. I've asked @tony-iqlusion to unyank the latest version until we get our fork updated.

steward/src/somm_send.rs
cbrit: Addressed

steward/src/allocation.rs
cbrit: Eric and I talked OOB, since clippy wants it this way for string interpolation let's leave it

steward/src/somm_send.rs
```diff
@@ -87,14 +87,22 @@ pub async fn data_hash(
...
         hasher.update(msg.as_bytes());
 
         return Ok(AllocationPrecommit {
             hash: hasher.finalize().to_vec(),
-            cellar_id: vote.cellar.clone().unwrap().id,
+            cellar_id: vote
+                .cellar
+                .clone()
+                .ok_or_else::<Error, _>(|| {
+                    ErrorKind::AllocationError
```
EricBolten: Shouldn't the presence of a cellar be checked at the time a vote is instantiated, as it would be invalid without it (e.g. it shouldn't be an `Option`)? Also, we'd want to check that the value is there before cloning. This would probably be cleaner pulling that value out before assigning it in the `return` statement.

steward/src/somm_send.rs
```diff
@@ -60,7 +60,7 @@ async fn __send_messages(
     fee: Coin,
     messages: Vec<Msg>,
 ) -> Result<TxResponse, CosmosGrpcError> {
-    let cosmos_address = cosmos_key.to_address(&contact.get_prefix()).unwrap();
+    let cosmos_address = cosmos_key.to_address(&contact.get_prefix())?;
```
EricBolten, line 63: We should still consider moving these repeated computations that are dependent on config values to config instantiation time.

EricBolten (commented):
Further comments.

steward/src/somm_send.rs
cbrit: Created #84 to cover this

steward/src/somm_send.rs
cbrit: It is an `Option<T>` because that's how prost generates the RebalanceVote struct field. Will move that code up before the assignment.

EricBolten (approved):
LGTM

Issue #36, opened 2021-12-16
Data Provider Payload/cellar_id Validation


cbrit:
Steward should make some kind of check that the gRPC method executed is an appropriate one to use against the provided cellar specified in the `cellar_id` field of the request. If it isn't, the request should not be processed and a response should be returned indicating failure with the relevant error code and message. 

For example, the `cellars.uniswapv3/Rebalance` method could feasibly be requested against a `cellar_id` matching the Aave cellar.

I think this would be an appropriate use for a Middleware/Layer in the tonic server?

cbrit:
Uniswap may not happen so we said we'd close this if not picked up in the next couple of weeks.

PR #83, opened 2022-02-24
Collin/cork

cbrit:
I think CI is going to fail because the integration test depends on a sommelier release that contains the cork module.


integration_tests/cork_test.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func ABIEncodedRebalance() []byte {
+	encodedCall, err := abi.JSON(strings.NewReader(CellarABISubset))
+	if err != nil {
+		panic(sdkerrors.Wrap(err, "bad ABI definition in code"))
+	}
+	args := []ABIEncodedTickRange{
+		{
+			TokenID: big.NewInt(int64(0)),
```
EricBolten: Does it matter whether or not the source is a `uint` here? Are the input int sizes not relevant when instantiating `big.NewInt`?

integration_tests/cork_test.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		panic(fmt.Sprintf("this is the error: %s", err))
+	}
+
+	return abiEncodedCall
+}
+
+func (s *IntegrationTestSuite) TestCork() {
+	s.Run("Bring up chain, and submit a cork", func() {
+		expectedTickRange := &allocationTypes.TickRange{
```
EricBolten, line 20: Nit: let's not meme our test values =P

integration_tests/cork_test.go
```diff
@@ -0,0 +1,257 @@
...
+`
+
+type ABIEncodedTickRange struct {
+	TokenID *big.Int `abi:"tokenId"`
+	Upper   *big.Int `abi:"tickUpper"`
+	Lower   *big.Int `abi:"tickLower"`
+	Weight  *big.Int `abi:"weight"`
+}
+
+func ABIEncodedRebalance() []byte {
```
EricBolten: On second look, is this function called at all?

integration_tests/setup_test.go
```diff
@@ -346,21 +349,30 @@ func (s *IntegrationTestSuite) initGenesis() {
...
 	allocationGenState.Params.VotePeriod = 30
 	bz, err = cdc.MarshalJSON(&allocationGenState)
 	s.Require().NoError(err)
-	appGenState[types.ModuleName] = bz
+	appGenState[allocationtypes.ModuleName] = bz
+
+	var corkGenState corktypes.GenesisState
+	s.Require().NoError(cdc.UnmarshalJSON(appGenState[corktypes.ModuleName], &corkGenState))
+	corkGenState.CellarIds = &corktypes.CellarIDSet{
+		Ids: []string{hardhatCellar.Hex()},
```
EricBolten, line 373: The cellar ID here is using `hardhatCellar.Hex()` and allocation is using `hardhatCellar.String()`. Is there a meaningful difference? Are we using different types on the chain side to represent these values?

steward/src/cork.rs
```diff
@@ -0,0 +1,107 @@
...
+        request: tonic::Request<SubmitRequest>,
+    ) -> Result<tonic::Response<SubmitResponse>, tonic::Status> {
+        debug!("received contract call request");
+        tokio::spawn(async move {
+            let request = request.get_ref();
+            let cork = match build_cork(request).await {
+                Ok(c) => c,
+                Err(err) => {
+                    error!("{}", err);
+                    return;
```
EricBolten: In these error cases, shouldn't we return a relevant `tonic::Status` indicating an error condition? Can we propagate a value like this out of the `tokio` thread?

steward/src/cork.rs
```diff
@@ -0,0 +1,107 @@
...
+                }
+            };
+            debug!("cork: {:?}", cork);
+
+            if let Err(err) = send_cork(cork).await {
```
EricBolten: Before sending, it would be useful to add a check to see if this is even a cellar that has been added by governance. The chain will also catch this error, but we can catch it too before submitting a vote by retrieving the values from the chain.

steward/src/cork.rs
```diff
@@ -0,0 +1,107 @@
...
+async fn send_cork(cork: Cork) -> Result<TxResponse, Error> {
+    let config = APP.config();
+    debug!("establishing grpc connection");
+    let timeout = Duration::from_secs(10);
+    let contact = Contact::new(&config.cosmos.grpc, timeout, &config.cosmos.prefix)?;
+
+    debug!("loading the delegate (orchestrator) key and address from config");
+    let name = &config.keys.rebalancer_key;
+    let delegate_key = config.load_deep_space_key(name.clone());
+    let delegate_address = delegate_key.to_address(&config.cosmos.prefix)?;
```
EricBolten: I think these are further examples of things we can load at config generation time as they won't change throughout the life of the instance.

steward/src/somm_send.rs
```diff
@@ -54,6 +61,21 @@ pub async fn send_allocation(
...
 
+pub async fn send_cork(
+    contact: &Contact,
+    cork: Cork,
+    delegate_address: String,
+    delegate_key: CosmosPrivateKey,
+    fee: Coin,
+) -> Result<TxResponse, CosmosGrpcError> {
+    let msg = MsgSubmitCorkRequest {
+        cork: Some(cork),
```
EricBolten, line 67: Why is this an `Option`? There is no valid `MsgSubmitCorkRequest` that doesn't contain a `cork`.

steward_proto/proto/steward.proto
```diff
@@ -0,0 +1,28 @@
...
+    string cellar_id = 1;
+    oneof contract_call_data {
+        UniswapV3RebalanceParams uniswapv3_rebalance = 2;
+    }
+}
+
+message SubmitResponse {}
+
+message UniswapV3Position {
+    int32 upper_price = 2;
```
EricBolten: Let's start this at id 1.

EricBolten (commented):
Some thoughts.

integration_tests/cork_test.go
cbrit: Haha, i had this in two places and only corrected it in one. Whoops.

integration_tests/cork_test.go
cbrit: Actually no. I put it here for debugging to make sure the encoded call matched what steward was getting, but then I stopped logging it once it was verified. I can remove it.

integration_tests/setup_test.go
cbrit: It looks like it's the same thing

https://github.com/ethereum/go-ethereum/blob/7231b3efb8095d3dd18d7164c3fa84d7705759d3/common/types.go#L234-L242

I think `String()` was added so that `Address` implements the `fmt.Stringer` interface according to the comments.

steward/src/cork.rs
cbrit: It looks like the cases where `build_cork` errors are when
1. The [cellar_id is improperly formatted](https://github.com/PeggyJV/steward/blob/da905a21c2ec8f892cc36a048184907510855f28/steward/src/cork.rs#L53) (address is wrong length, there is no ':', etc.)
2. The [contract call data is missing](https://github.com/PeggyJV/steward/blob/da905a21c2ec8f892cc36a048184907510855f28/steward/src/cork.rs#L55-L58)
3. The [cellar address is not whitelisted in the steward config](https://github.com/PeggyJV/steward/blob/da905a21c2ec8f892cc36a048184907510855f28/steward/src/cork.rs#L75)

I can see all of these being covered by an `InvalidArguments` code wrapped in the `tonic::Status`. Do you think that is appropriate?

It is possible that if 3) occurs it's not because the argument is invalid but because the validator failed to update their config appropriately, but that would be made clear by the internal error message.

steward/src/cork.rs
cbrit: #84 should cover this

steward/src/somm_send.rs
cbrit: I agree. This is on the sommelier proto side and is how prost generates the message type for some reason.

steward/src/cork.rs
```diff
@@ -0,0 +1,144 @@
...
+                ),
+            ));
+        }
+
+        // Build and send cork
+        let cork = match build_cork(request).await {
+            Ok(c) => c,
+            Err(err) => {
+                warn!("failed to build cork: {}", err);
+                return Err(Status::new(Code::InvalidArgument, err.to_string()));
```
brittcoin, line 74: Unsure what this will looking like in the error logs if the strategy provider. `err.to_string()` might result in a string of nested rust error types like 

```
AnError(ANestedError("error message")):100:13
```

cbrit, line 78: Unsure how this will end up looking in the strategy provider's logs. `err.to_string()` may result in a string of nested rust error types like 
```
SomeError(NestedError("error message")):100:13
```
Would be easy test error behavior with a sommelier grpc mock service.
I don't think it's a huge deal right now personally but something we may want to improve later.

steward/src/cork.rs
EricBolten: The first two are cases of invalid arguments, the third I'd say is an authorization failure.

steward/src/cork.rs
```diff
@@ -0,0 +1,144 @@
...
+                error!("cork query client connection failed: {}", err);
+                return Err(Status::new(
+                    Code::Internal,
+                    "failed to query chain to validate cellar id",
+                ));
+            }
+        };
+        debug!("checking if cellar ID is approved");
+        let ids = &client
+            .query_cellar_i_ds(QueryCellarIDsRequest {})
```
EricBolten, line 54: I think ideally we have a background thread running to update this state so we don't have to make this call every request.

integration_tests/cork_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"io/ioutil"
+	"time"
+
+	allocationTypes "github.com/peggyjv/sommelier/v3/x/allocation/types"
+	corkTypes "github.com/peggyjv/sommelier/v3/x/cork/types"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+func (s *IntegrationTestSuite) TestCork() {
```
EricBolten, line 18: Let's add this test to CI on this branch.

steward/src/cellars.rs
```diff
@@ -17,6 +18,12 @@ impl std::fmt::Display for CellarId {
     }
 }
 
+#[derive(Debug, Deserialize)]
+pub struct CellarContractCall {
```
EricBolten: Is this being used anywhere?

steward/src/cork.rs
EricBolten: You'd have to check for the error types within `build_cork` and return an error formatted for the end user if you want this to look cleaner.

steward/src/cork.rs
```diff
@@ -0,0 +1,144 @@
...
+            Err(err) => {
+                warn!("failed to build cork: {}", err);
+                return Err(Status::new(Code::InvalidArgument, err.to_string()));
+            }
+        };
+        debug!("cork: {:?}", cork);
+
+        if let Err(err) = send_cork(cork).await {
+            error!("failed to submit cork: {}", err);
+            return Err(Status::new(Code::Internal, "failed to submit transaction"));
```
EricBolten: This should specify that Steward failed to send the cork to the chain (lacking detail it could also just sound like the request to Steward failed).

steward/src/cork.rs
```diff
@@ -0,0 +1,144 @@
...
+                Err("cellar not found in config".to_string().into())
+            }
+        }
+    }
+}
+
+async fn send_cork(cork: Cork) -> Result<TxResponse, Error> {
+    let config = APP.config();
+    debug!("establishing grpc connection");
+    let timeout = Duration::from_secs(10);
```
EricBolten: This could be moved to a constant.

EricBolten (commented):
Some further comments.

steward/src/cellars.rs
cbrit: no

hannydevelop (commented):
Of what use is the LGTM analysis?

cbrit:
Closes #36

cbrit:
Closes #84

integration_tests/setup_test.go
```diff
@@ -60,7 +61,7 @@ func MNEMONICS() []string {
 var (
 	stakeAmount, _  = sdk.NewIntFromString("100000000000")
 	stakeAmountCoin = sdk.NewCoin(bondDenom, stakeAmount)
-	hardhatCellar   = common.HexToAddress("0x4C4a2f8c81640e47606d3fd77B353E87Ba015584")
+	hardhatCellar   = common.HexToAddress("0x0000000000000000000000000000000000000000")
```
EricBolten, line 66: Why is this address zeroed out?

steward/src/cellars/aave_v2_stablecoin.rs
```diff
@@ -0,0 +1,77 @@
...
+            let call = SetInputTokenCall {
+                token: params
+                    .address
+                    .parse::<H160>()
+                    .expect("failed to parse token address"),
+                is_approved: params.is_approved,
+            };
+            AaveV2StablecoinCellarCalls::SetInputToken(call).encode()
+        }
+        RemoveLiquidityRestriction(_) => {
```
EricBolten, line 63: Should this call be a strategy provider thing or a valset thing?

steward/src/config.rs
```diff
@@ -255,3 +273,9 @@ impl Default for MetricsSection {
         }
     }
 }
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct UniswapV3CellarInfo {
```
EricBolten: Is this being used anywhere?

steward/src/cork.rs
```diff
@@ -0,0 +1,145 @@
...
+
+        // Check if cellar is governance approved before building cork
+        let config = APP.config();
+        let mut client = match CorkQueryClient::connect(config.cosmos.grpc.clone()).await {
+            Ok(c) => c,
+            Err(err) => {
+                error!("cork query client connection failed: {}", err);
+                return Err(Status::new(
+                    Code::Internal,
+                    "failed to query chain to validate cellar id",
```
EricBolten, line 47: This error seems different from the one we're printing locally, this would just be the query connection failing to be created at all.

steward/src/cork.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+async fn send_cork(cork: Cork) -> Result<TxResponse, Error> {
+    let config = APP.config();
+    debug!("establishing grpc connection");
+    let contact = Contact::new(&config.cosmos.grpc, MESSAGE_TIMEOUT, &config.cosmos.prefix)?;
+
+    debug!("loading the delegate (orchestrator) key and address from config");
+    let delegate_key = &config::DELEGATE_KEY;
+    let delegate_address = &config::DELEGATE_ADDRESS;
```
EricBolten: We don't really need to assign these two to variables since we're only using them once.

steward_proto/proto/steward.proto
```diff
@@ -0,0 +1,32 @@
...
+import "aave_v2_stablecoin.proto";
+
+service ContractCall {
+    rpc Submit(SubmitRequest) returns (SubmitResponse) {}
+}
+
+message SubmitRequest {
+    string cellar_id = 1;
+    oneof call_data {
+        AaveV2Stablecoin aave_v2_stablecoin = 2;
```
EricBolten, line 15: If we're moving Uniswap to 100 presumably to get it out of the way to later be deleted, this should probably start at 1.

EricBolten (commented):
Some comments.

integration_tests/setup_test.go
cbrit: It could just be a defaulted string. `hardhatCellar` gets dynamically set by parsing the logs for whatever the address is now just like the gravity contract.

steward/src/cellars/aave_v2_stablecoin.rs
cbrit: That's a question I've been asking for a couple of weeks. I don't know which ones are/aren't for SPs and it's not documented.

steward/src/cork.rs
cbrit: I wasn't sure it was pertinent information to the SP what caused the query failure. Can change it if you think I should.

steward_proto/proto/steward.proto
cbrit: `cellar_id` is 1, or does `oneof` start its own count?

EricBolten: Whoops, totally missed that, no.

steward/src/cork.rs
EricBolten: Eh, not a huge deal.

integration_tests/setup_test.go
EricBolten: Mmm, I see that happens for gravity contract too. Might be worth a comment.

EricBolten (approved):
LGTM, I think it makes sense to merge this now to get this long-lived PR closed out and we can address the teardown issues in the follow-up testing PR.

PR #92, opened 2022-03-07
Remove incomplete commands 

### LINKED ISSUE #89: Incomplete CLI command cleanup
There are a variety of CLI commands that are either incomplete or half-implemented. The most expedient thing to do for now would be to simply delete them. We can always implement further commands at a later date, and having half-finished code is cumbersome and confusing.

hannydevelop:
Closes #89


steward/src/commands/tx/eth.rs
```diff
@@ -16,7 +16,6 @@ use gravity_bridge::gravity_utils::ethereum::{downcast_to_u64, format_eth_addres
 #[derive(Command, Debug, Parser)]
 pub enum Eth {
     SendToCosmos(SendToCosmos),
-    Send(Send),
 }
 
 impl Runnable for Eth {
```
cbrit, line 21: Nit: could we move empty `Runnable` impls into the `Eth` derive directive?

```
#[derive(Command, Debug, Parser, Runnable)]
pub enum Eth {
...
```

similarly with the `Cosmos` command

cbrit (approved):
Happy to see steward getting into shape!

PR #95, opened 2022-03-11
Remove dead code

### LINKED ISSUE #88: Dead code cleanup
Go through the repo and delete any code no longer relevant to the current iteration of steward.

hannydevelop:
Closes #88


steward/src/commands.rs
```diff
@@ -12,25 +12,19 @@
...
 mod allow_erc20;
 mod config_cmd;
-mod cosmos_mode;
 mod cosmos_to_eth;
 mod deploy;
 mod eth_to_cosmos;
-mod fund_cellar;
 mod keys;
 mod orchestrator;
-mod query;
```
cbrit, line 22: Looks like `query.rs` needs to be deleted since we're removing this

steward/src/cork.rs
```diff
@@ -108,13 +198,14 @@ async fn build_cork(request: &SubmitRequest) -> Result<Cork, Error> {
...
 
+#[allow(unreachable_patterns)]
 fn get_encoded_call(data: CallData) -> Result<Vec<u8>, Error> {
     match data {
         AaveV2Stablecoin(call) => Ok(aave_v2_stablecoin::get_encoded_call(
             call.function
                 .expect("call data for Aave V2 Stablecoin cellar was empty"),
         )),
-        Uniswapv3Rebalance(params) => Ok(uniswapv3::get_encoded_call(params)),
+        _ => Err(ErrorKind::Http.context("empty call data for cellar").into()),
```
cbrit: If this causes a warning that requires `#[allow(unreachable_patterns)]` to silence it, I think we probably don't need this variant.

steward_abi_build/src/main.rs
```diff
@@ -137,62 +85,4 @@ fn main() {
         Ok(_) => (),
         Err(e) => println!("Error writing weth.rs: {}", e),
     }
```
cbrit: I'm not actually sure what this is for. Do we need it? My guess is no.

steward/src/cork.rs
```diff
@@ -93,6 +99,90 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
...
+
+        let address: H160 = cellar_address
+            .parse()
+            .expect("could not parse cellar address");
+
+        let contract_call_data = match request.call_data.clone() {
+            Some(call) => call,
+            None => return Err(tonic::Status::invalid_argument("err")),
+        };
+        let encoded_call = get_encoded_call(contract_call_data).unwrap();
```
cbrit: Since we aren't corking the call data we don't have to encode it ourselves and can use the convenience methods like we did with those uniswap commands. Something like this (possibly refactored into methods..):

```rust
use steward_abi::aave_v2_stablecoin::AaveV2StablecoinCellar;
// ...
match contract_call_data {
    AaveV2Stablecoin(call) => {
        let contract = AaveV2StablecoinCellar::new(address, Arc::new(client));

        match call.function.unwrap() {
            AccruePlatformFees(_) => contract.accrue_platform_fees(),
            // ... etc for other functions
        }
    }
};
```

cbrit (commented):
Excited to see Steward losing some weight!

cbrit:
Oh and I forgot! we also don't need the CellarConfig config section vector anymore.

steward/src/commands.rs
hannydevelop: Ok

steward/src/cork.rs
hannydevelop: Removing that variant will break the code. It's a bad practice to have one variant in an Enum. However, since we have only one cellar, there's nothing I can do than try to patch things up.

That warning originates because the compiler can't reach the catch-all `(_)` arm. That's because the match section won't run into another variant. The reason is because we have one cellar which is `Aavestablecoin` and we're telling the match section that `Aavestablecoin` call is `Ok`. Since we'll only get `Aavestablecoin` call for now, we can't run into another variant.

https://www.google.com/url?sa=t&source=web&rct=j&url=https://doc.rust-lang.org/book/ch06-02-match.html%23:~:text%3DRust%2520also%2520has%2520a%2520pattern,us%2520about%2520an%2520unused%2520variable.&ved=2ahUKEwjZ24fc45f3AhXIgv0HHSgsDT0QFnoECA8QBQ&usg=AOvVaw1ZAHU85GqibST4ycgRtnUo

While the catch-all match arm is useless, removing it will throw an error, since we can't have one match arm in a match pattern

steward_abi_build/src/main.rs
hannydevelop: Weth's not need then, I'll take it out

steward/src/cork.rs
hannydevelop: This will work!

hannydevelop:
> 

Really? That's a relief

steward/src/cork.rs
hannydevelop: In my opinion, we'll remove this once we have another cellar. However, I'm open to suggestions on how to fix this if it'll cause a breach of security or break anything.

cbrit: > Removing that variant will break the code. It's a bad practice to have one variant in an Enum. However, since we have only one cellar, there's nothing I can do than try to patch things up.

Removing the `_` variant doesn't break anything, I just tested it.

hannydevelop: Damn! It kept saying "can't have a single match arm" until I ran cargo clean

cbrit: Weird. I wonder if there was an old version of the enum in the cache or something.

steward/src/cork.rs
```diff
@@ -93,6 +99,90 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
...
+            downcast_to_u64(chain_id).expect("Chain ID overflowed when downcasting to u64");
+        let client = SignerMiddleware::new(provider, wallet.clone().with_chain_id(chain_id));
+
+        let cellar_address = request.cellar_id.clone();
+
+        if let Err(err) = cellars::validate_cellar_id(&cellar_address) {
+            return Err(tonic::Status::invalid_argument(err));
+        }
+
+        let address: H160 = cellar_address
```
EricBolten: We should use the ethers `Address` type specifically, rather than referring to addresses as `H160` which is just a sized hash type.

steward/src/cork.rs
```diff
@@ -93,6 +99,90 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
...
+
+            // create the contract object at the address
+            let contract = Contract::new(address, abi, &client);
+
+            // Calling constant methods is done by calling `call()` on the method builder.
+            // (if the function takes no arguments, then you must use `()` as the argument)
+            contract
+                .method::<_, String>("getValue", ())
+                .unwrap()
+                .call()
```
EricBolten: `call` works for non-mutating state contract calls. In order to broadcast a transaction you would have to use `send`, which has some more steps involved. Here's an example from the gravity bridge, though it's wrapping it in a timeout:

https://github.com/PeggyJV/gravity-bridge/blob/fb045a1867b723123fdc54d01f8e8be936a7189e/orchestrator/ethereum_gravity/src/logic_call.rs#L64-L90

steward/src/cork.rs
EricBolten: Agreed, let's use the convenience methods for now.

EricBolten (commented):
A few comments.

EricBolten (approved):
LGTM

Issue #98, opened 2022-03-24
Create mock Aave contract for steward testing


cbrit:
Cellar contract that simply emits an event or sets some value for testing with no dependencies.

PR #99, opened 2022-03-28
replace test cellar with aave v2 stabelcoin mock

cbrit:
I've only mocked the methods that steward exposes for strategy providers. In the future we'll probably want to fill it out more to support scheduled cork commands.


cbrit:
Closes #98

go.mod
```diff
@@ -5,14 +5,16 @@ go 1.15
...
-	github.com/ethereum/go-ethereum v1.10.11
+	github.com/ethereum/go-ethereum v1.10.16
 	github.com/miguelmota/go-ethereum-hdwallet v0.1.1
 	github.com/ory/dockertest/v3 v3.8.1
 	github.com/peggyjv/gravity-bridge/module v0.3.9
 	github.com/peggyjv/sommelier/v3 v3.1.2-0.20220318164959-6feb04a85567
 	github.com/spf13/viper v1.8.1
 	github.com/stretchr/testify v1.7.0
 	github.com/tendermint/tendermint v0.34.14
+	golang.org/x/crypto v0.0.0-20220321153916-2c7772ba3064 // indirect
```
EricBolten: What caused these go.mod updates? Are they necessary?

integration_tests/aave_abi.go
```diff
@@ -0,0 +1,1757 @@
+// Code generated - DO NOT EDIT.
```
EricBolten, line 1: How was this generated? Also wondering if it actually gets referenced anywhere, since it seems like the reference to functions calls/topics are generated by hand in `cork_test`.

integration_tests/cork_test.go
```diff
@@ -154,27 +147,87 @@ func (s *IntegrationTestSuite) TestCork() {
...
-				s.T().Logf("wrong upper %s", actualTickRange.String())
-				return false
+			logs, err := ethClient.FilterLogs(context.Background(), query)
+			s.Require().NoError(err)
+			s.T().Logf("got %v logs", len(logs))
+
+			// For non-anonymous events, the first log topic is a keccak256 hash of the
+			// event signature.
+			eventSignature := []byte("LogicCallEvent(bytes32,uint256,bytes,uint256)")
+			signatureTopic := crypto.Keccak256Hash(eventSignature).Hex()
```
EricBolten: Is this something we could pull out of a generated ABI similar to what it looks like was done with the Aave contract here?

integration_tests/ethereum/contracts/MockAaveV2StablecoinCellar.sol
```diff
@@ -0,0 +1,113 @@
...
+    function accrueFees() external {
+        emit mockAccrueFees();
+    }
+
+    event mockTransferFees();
+
+    /**
+     * @notice Transfer accrued fees to Cosmos to distribute.
+     */
+    function transferFees() external  {
```
EricBolten: A lot of functions (e.g. this one) are defined in the actual contract as `onlyOwner`. Is this going to have a material impact on the test conditions?

EricBolten (commented):
A few questions.

integration_tests/aave_abi.go
cbrit: Artifact from when I was trying to test for the event a different way. Will delete.

integration_tests/ethereum/contracts/MockAaveV2StablecoinCellar.sol
cbrit: It will have no bearing on the mock contract in its current state.

integration_tests/cork_test.go
cbrit: Possibly, I'll check!

cbrit: AFAICT you you can use log filtering directly from the generated ABI but the objects you need can only be obtained if you also deploy the contract with go-ethereum. Since we are deploying with hardhat I don't think it's an option.

integration_tests/ethereum/contracts/MockAaveV2StablecoinCellar.sol
EricBolten: Is the reason we didn't include it because there isn't a stable owner address?

cbrit: I don't think it was a conscious decision, I must have removed it by mistake

cbrit: There actually should be because we are changing the owner in the hardhat config. I'll re-add and try it out. I misunderstood your comment as referring to `transferFees` and not the functions in general

integration_tests/ethereum/contracts/mock_aave_abi.json
```diff
@@ -0,0 +1,230 @@
...
+			}
+		],
+		"name": "sweep",
+		"outputs": [],
+		"stateMutability": "nonpayable",
+		"type": "function"
+	},
+	{
+		"inputs": [],
+		"name": "transferFees",
```
EricBolten: Does this ABI need to be updated as well since the contract changed to include onlyOwner, or is it not part of this signature?

integration_tests/cork_test.go
```diff
@@ -154,27 +147,87 @@ func (s *IntegrationTestSuite) TestCork() {
...
+
+			for _, vLog := range logs {
+				var topics [4]string
+				for i := range vLog.Topics {
+					topics[i] = vLog.Topics[i].Hex()
+				}
+
+				s.T().Logf("found signature topic: %s...", topics[0][:10])
+				if topics[0] == signatureTopic {
+					s.T().Logf("gravity logic call event found!")
```
EricBolten: Should we be verifying the contents of the event match the call we're looking for? We're just verifying the presence of one on that topic.

EricBolten (commented):
A couple comments.

integration_tests/ethereum/contracts/mock_aave_abi.json
cbrit: Cross-referencing other ABI JSONs, it doesn't look like modifiers are included in the function definitions. They definitely are not included in the signature as far as encoding goes.

integration_tests/cork_test.go
cbrit: Probably a good idea... It is the only `submitLogicCall` call occurring in the test but could cause weird false positives in the future if that changes.

EricBolten:
Seems like the test is also busted atm for some reason, worth looking into.

cbrit:
> Seems like the test is also busted atm for some reason, worth looking into.

It's a transient bug with the eth client that I have yet to solve.

integration_tests/cork_test.go
```diff
@@ -154,27 +150,134 @@ func (s *IntegrationTestSuite) TestCork() {
...
+
+					event := &GravityLogicCallEvent{
+						InvalidationId:    out[0].([32]byte),
+						InvalidationNonce: out[1].(*big.Int),
+						ReturnData:        out[2].([]byte),
+						EventNonce:        out[3].(*big.Int),
+					}
+					s.T().Logf("log%v: comparing invalidation parameters", i)
+					eventInvalidationId := event.InvalidationId[:]
+					if reflect.DeepEqual(eventInvalidationId, invalidationScope) && int(event.InvalidationNonce.Int64()) == invalidationNonce {
```
EricBolten: I think you can use `bytes.Compare` here for byte slices?

integration_tests/cork_test.go
```diff
@@ -154,27 +150,134 @@ func (s *IntegrationTestSuite) TestCork() {
...
+			s.Require().NoError(err)
+			s.T().Logf("got %v gravity.submitLogicCall logs", len(logs))
+			ethClient.Close()
+
+			gravity_abi, err := GravityMetaData.GetAbi()
+			s.Require().NoError(err)
+
+			for i, log := range logs {
+				if len(log.Data) > 0 {
+					out, err := gravity_abi.Unpack("LogicCallEvent", log.Data)
```
EricBolten: I think you can use `UnpackIntoInterface` and pass it a pointer to an empty struct.

integration_tests/cork_test.go
```diff
@@ -154,27 +150,134 @@ func (s *IntegrationTestSuite) TestCork() {
...
+					},
+				},
+			}
+
+			logs, err := ethClient.FilterLogs(context.Background(), query)
+			s.Require().NoError(err)
+			s.T().Logf("got %v logs", len(logs))
+			ethClient.Close()
+
+			if len(logs) > 0 {
```
EricBolten: In this particular test I think we would expect `len(logs) == 1`.

integration_tests/cork_test.go
```diff
@@ -154,27 +150,134 @@ func (s *IntegrationTestSuite) TestCork() {
...
-				return false
+			logs, err := ethClient.FilterLogs(context.Background(), query)
+			s.Require().NoError(err)
+			s.T().Logf("got %v gravity.submitLogicCall logs", len(logs))
+			ethClient.Close()
+
+			gravity_abi, err := GravityMetaData.GetAbi()
+			s.Require().NoError(err)
+
+			for i, log := range logs {
```
EricBolten: Similarly here, I think we would only expect a single event to be returned on the topic for logic call events.

EricBolten (commented):
A few comments.

EricBolten (approved):
LGTM

PR #101, opened 2022-04-06
Steward v3: Cork V2 support and cellar ID caching

### LINKED ISSUE #93: Background thread maintaining state of governance-approved cellar IDs
As things are currently written (as of now in the branch `collin/cork`), each time a strategy provider sends data to steward will incur a gRPC query to the cork module to check if the contained cellar_id is approved by governance. To reduce overhead, we should maintain a local copy of this state that updates periodically. If data is submitted to steward with a cellar_id not contained in steward's local state, it should either then query the cork module to make sure it hasn't been added since the last time steward updated it's state, or just force a state update and then check again.

cbrit:
Closes #93


steward/src/commands/cosmos_mode.rs
```diff
@@ -26,6 +34,16 @@ impl Runnable for CosmosSignerCmd {
         let config = APP.config();
         info!("Starting application");
         abscissa_tokio::run(&APP, async {
+            if let Err(err) = CORK_QUERY_CLIENT.set(CorkClientWrapper::new().await.unwrap()) {
+                // SetError<T> does not implement Debug and therefore can't call unwrap() or expect()
+                panic!("{}", err);
+            }
```
tony-iqlusion: Wow, that's a really weird deficiency in `SetError`

cbrit: right?

cbrit:
Need to make the cache a `HashSet<T>` instead of a vector

EricBolten:
This one is still pending merge conflict resolution.

EricBolten:
Another ping here, an additional merge conflict after the ABI update.

EricBolten:
I see this has been moved into draft -- is it still a work in progress or ready for review?

go.mod
```diff
@@ -9,7 +9,8 @@ require (
 	github.com/miguelmota/go-ethereum-hdwallet v0.1.1
 	github.com/ory/dockertest/v3 v3.8.1
 	github.com/peggyjv/gravity-bridge/module/v2 v2.0.2
-	github.com/peggyjv/sommelier/v4 v4.0.2
+	// TO-DO: Change back when cork v2 is merge in sommelier
```
EricBolten: Nit: to make it searchable and consistent, prefer using `TODO`.

integration_tests/cork_test.go
```diff
@@ -383,27 +403,16 @@ func (s *IntegrationTestSuite) checkCellarExists(cellar common.Address) {
...
-			}
-			return false
-		}
-
-		return true
-	}, 30*time.Second, 1*time.Second, "new vote period never seen")
+		currentHeight, err := s.GetLatestBlockHeight(clientCtx)
+		s.Require().NoError(err)
+		return currentHeight >= height
+	}, 30*time.Second, 1*time.Second, "scheduled height never reached")
```
EricBolten: Nit: picking a 30 second timeout to reach an arbitrary height seems brittle. Not sure if there's a better solution necessarily, perhaps `block height delta at call * 2 seconds * 2` or something.

integration_tests/proposal_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+	orch := s.chain.orchestrators[0]
+	orchClientCtx, err := s.chain.clientContext("tcp://localhost:26657", orch.keyring, "orch", orch.keyInfo.GetAddress())
+	s.Require().NoError(err)
+	currentHeight, err := s.GetLatestBlockHeight(orchClientCtx)
+	s.Require().NoError(err)
+	protoJson := `
+	{
+		"call": {
+			"CellarV1": {
```
EricBolten, line 30: Shouldn't this be `CellarV1Governance`? Why does this work?

integration_tests/proposal_test.go
```diff
@@ -0,0 +1,185 @@
...
+						"position": {
+							"Erc20Address": "0x0000000000000000000000000000000000000000"
+						}
+					}
+				}
+			}
+		}
+	}
+	`
+	targetBlockHeight := currentHeight + 120
```
EricBolten, line 42: I'm surprised we hit 120 blocks waiting only 30 seconds in the `waitForScheduledHeight` call.

integration_tests/proposal_test.go
```diff
@@ -0,0 +1,185 @@
...
+	s.T().Log("Proposal approved!")
+
+	s.T().Log("Waiting for scheduled cork to be created by steward")
+	corkQueryClient := corktypes.NewQueryClient(orchClientCtx)
+	s.Require().Eventually(func() bool {
+		proposalQueryResponse, _ := corkQueryClient.QueryScheduledCorks(context.Background(), &corktypes.QueryScheduledCorksRequest{})
+		return len(proposalQueryResponse.Corks) > 0
+	}, time.Second*120, time.Second*2, "corks never scheduled")
+
+	s.T().Log("wait for scheduled height")
```
EricBolten, line 117: Oh, I see, we're not using that utility function here.

proto/cellar_v1.proto
```diff
@@ -228,3 +228,89 @@ message CellarV1 {
...
+        // Represents function `initiateShutdown()`
+        InitiateShutdown initiate_shutdown = 1;
+        // Represents function `liftShutdown()`
+        LiftShutdown lift_shutdown = 2;
+        // Represents function `resetHighWatermark()`
+        ResetHighWatermark reset_high_watermark = 3;
+        // Represents function `setFeesDistributor(address)`
+        SetFeesDistributor set_fees_distributor = 4;
+        // Represents function `setOwner(address)`
+        SetOwner set_owner = 5;
```
EricBolten: I think we left `TransferOwnership` out of the Aave cellar. Should we drop this, or was our initial thought wrong and maybe it's best to include both? It's still technically possible to compose a call to do this without supporting it, just manual and difficult.

integration_tests/proposal_test.go
```diff
@@ -0,0 +1,185 @@
...
+	s.checkCellarExists(vaultCellar)
+
+	orch := s.chain.orchestrators[0]
+	orchClientCtx, err := s.chain.clientContext("tcp://localhost:26657", orch.keyring, "orch", orch.keyInfo.GetAddress())
+	s.Require().NoError(err)
+	currentHeight, err := s.GetLatestBlockHeight(orchClientCtx)
+	s.Require().NoError(err)
+	protoJson := `
+	{
+		"call": {
```
EricBolten, line 29: Should the top-level message in here be an instance of `GovernanceCall` or is it implied that the contents represent one?

proto/steward.proto
```diff
@@ -17,21 +17,23 @@ import "cellar_v1.proto";
...
+message ScheduleRequest {
     // The ID (currently simply an Ethereum address) of the target Cellar
     string cellar_id = 1;
     // The data from which the desired contract function will be encoded
     oneof call_data {
         AaveV2Stablecoin aave_v2_stablecoin = 2;
         CellarV1 cellar_v1 = 3;
     }
+    // The block height at which to schedule the contract call
+    uint64 block_height = 4;
```
EricBolten: Since we're doing an API-breaking change, I suggest we move `block_height` to ID 2 so that `call_data` can grow sequentially.

steward/src/cellars/aave_v2_stablecoin.rs
```diff
@@ -202,3 +208,97 @@ fn validate_route(results: Vec<Result<Address, &String>>) -> Result<(), Error> {
...
+) -> Result<Vec<u8>, Error> {
+    match function {
+        SetFeesDistributor(params) => {
+            log_governance_cellar_call(
+                proposal_id,
+                CELLAR_NAME,
+                &SetFeesDistributorCall::function_name(),
+                cellar_id,
+            );
+            let address: CosmosAddress = params.new_fees_distributor.parse().map_err(|err| {
```
EricBolten, line 227: `SetFeesDistributor` takes a `bytes32` input, which is the padded to 32 length value of the raw 20 byte Cosmos address representation. If we're taking a Cosmos address as input, we'll have to perform this conversion.

You can see an example of the expected data formatting here: https://github.com/PeggyJV/gravity-bridge/blob/4a107a812e4c5afd10ba795b016f6558b206e5f8/orchestrator/ethereum_gravity/src/send_to_cosmos.rs#L43-L54

steward/src/cellars/aave_v2_stablecoin.rs
```diff
@@ -202,3 +208,97 @@ fn validate_route(results: Vec<Result<Address, &String>>) -> Result<(), Error> {
...
+            log_governance_cellar_call(
+                proposal_id,
+                CELLAR_NAME,
+                &SetTrustCall::function_name(),
+                cellar_id,
+            );
+            let call = SetTrustCall {
+                trust: params.trust,
+                position: params.position.parse::<Address>().map_err(|err| {
+                    sp_call_error(format!(
```
EricBolten: Similar to the logging, should governance errors have their own function rather than being `sp_call_error`?

steward/src/cellars/cellar_v1.rs
```diff
@@ -230,3 +236,166 @@ fn encode_swap_params(params: SwapParams) -> Result<Vec<u8>, Error> {
...
+) -> Result<Vec<u8>, Error> {
+    match function {
+        SetFeesDistributor(params) => {
+            log_governance_cellar_call(
+                proposal_id,
+                CELLAR_NAME,
+                &SetFeesDistributorCall::function_name(),
+                cellar_id,
+            );
+            let address: CosmosAddress = params.new_fees_distributor.parse().map_err(|err| {
```
EricBolten, line 253: Same here for fees distributor.

steward/src/cellars/cellar_v1.rs
```diff
@@ -230,3 +236,166 @@ fn encode_swap_params(params: SwapParams) -> Result<Vec<u8>, Error> {
...
+                    LOG_PREFIX
+                )));
+            }
+            let position = params.position.unwrap();
+            let position_address: String;
+            // a little weird but necessary to avoid adding an overly complex enum to the proto
+            let position_type: u8 = match position {
+                Position::Erc20Address(address) => {
+                    position_address = address;
+                    0
```
EricBolten, line 357: Weird, is there not a way to retrieve the integer enum value without manually doing it?

steward/src/commands/cork_proposal.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+/// Outputs the JSON formatted scheduled cork data for submitting a Scheduled Cork Proposal to Sommelier
+fn print_proposal(height: u64, contract: String, governance_call: GovernanceCall, quiet: bool) {
+    let json =
+        serde_json::to_string(&governance_call).expect("failed to serialize governance call");
+    let proposal = ScheduledCorkProposal {
+        block_height: height,
+        target_contract_address: contract,
+        contract_call_proto_json: json,
```
EricBolten, line 29: I assume this is correctly escaping the interior JSON then?

steward/src/commands/cork_proposal/aave_v2_stablecoin/sweep.rs
```diff
@@ -0,0 +1,68 @@
...
+                .await
+                .unwrap_or_else(|err| {
+                    status_err!("cellar ID validation error:: {}", err);
+                    std::process::exit(1);
+                });
+
+            let governance_call = GovernanceCall {
+                call: Some(Call::AaveV2Stablecoin(AaveV2StablecoinGovernance {
+                    function: Some(Function::Sweep(Sweep {
+                        token: format!("{:#x}", self.token),
```
EricBolten, line 50: `ethers` has issues with printing addresses, I assume you noticed their weird truncation and that this implementation works? Maybe this is better than the jank-ass way it's done in the orchestrator atm: https://github.com/PeggyJV/gravity-bridge/blob/4a107a812e4c5afd10ba795b016f6558b206e5f8/orchestrator/gravity_utils/src/ethereum.rs#L39-L41

steward/src/commands/cork_proposal/aave_v2_stablecoin/set_fees_distributor.rs
```diff
@@ -0,0 +1,63 @@
...
+
+/// Fees Distributor subcommand
+#[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION\n\nSet fees distributor Cosmos address of target cellar when chain reaches specified height"
+)]
+pub struct FeesDistributorCmd {
+    #[clap(short, long)]
+    /// Fee distributor's address
+    new_fees_distributor: Address,
```
EricBolten: Nit: it's a tad confusing that all the prop commands import `Address` but in some it means a Cosmos address and in some it means an Ethereum address. It might be clearer to use a named import, e.g. `as CosmosAddress`.

steward/src/commands/cork_proposal/aave_v2_stablecoin/sweep.rs
```diff
@@ -0,0 +1,68 @@
...
+use steward_proto::steward::{
+    aave_v2_stablecoin_governance::{Function, Sweep},
+    governance_call::Call,
+    AaveV2StablecoinGovernance, GovernanceCall,
+};
+
+/// Sweep subcommand
+#[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION\n\nSchedule sweep Cosmos address of target cellar when chain reaches specified height"
```
EricBolten: If we're linking to the contract in the `CellarV1` commands, should we do it for Aave as well?

steward/src/commands/cork_proposal/cellar_v1/initiate_shutdown.rs
```diff
@@ -0,0 +1,55 @@
...
+use abscissa_core::{clap::Parser, Command, Runnable};
+use steward_proto::steward::{
+    cellar_v1_governance::{Function, InitiateShutdown},
+    governance_call::Call,
+    CellarV1Governance, GovernanceCall,
+};
+
+#[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION\n\nCalls initiateShutdown() on the target cellar contract at the specified block height.\nFor more information see https://github.com/PeggyJV/cellar-contracts/blob/main/src/base/Cellar.sol"
```
EricBolten: `cellar-contracts` is going to be pointing at Cellar V2. You probably want https://github.com/PeggyJV/cellar-v1_5/blob/release/src/base/Cellar.sol

steward/src/commands/cork_proposal/cellar_v1/set_fees_distributor.rs
```diff
@@ -0,0 +1,68 @@
...
+    fn run(&self) {
+        abscissa_tokio::run_with_actix(&APP, async {
+            cellars::validate_cellar_id(&self.cellar_id)
+                .await
+                .unwrap_or_else(|err| {
+                    status_err!("invalid cellar ID: {}", err);
+                    std::process::exit(1);
+                });
+
+            if self.new_fees_distributor.get_prefix().ne("somm") {
```
EricBolten, line 43: Nit: I believe the full qualified prefix is `somm1`, which indicates the end of the prefix.

steward/src/cork.rs
```diff
@@ -11,73 +14,57 @@ use abscissa_core::{
...
-                    &request.cellar_id
-                ),
-            ));
         }
 
         // Build and send cork
         let cellar_id = request.cellar_id.clone();
-        let cork = match build_cork(request).await {
+        let height = request.block_height;
+        if let Err(err) = cellars::validate_cellar_id(&request.cellar_id).await {
```
EricBolten: `validate_cellar_id` was already called above.

steward/src/cork/cache.rs
```diff
@@ -0,0 +1,69 @@
...
+    static ref APPROVED_CELLARS: ApprovedCellarsCache = ApprovedCellarsCache::default();
+}
+
+/// Indicates whether an address is in the approved cellars cache
+pub fn is_approved(cellar_id: &str) -> bool {
+    let cellar_id = cellar_id.trim().to_lowercase();
+    APPROVED_CELLARS
+        .read()
+        .unwrap()
+        .iter()
```
EricBolten: Given that it's a `HashSet`, shouldn't lookup not require iteration?

steward/src/cork/client.rs
```diff
@@ -0,0 +1,27 @@
+use crate::{error::Error, prelude::APP};
+use abscissa_core::Application;
+use somm_proto::cork::{query_client::QueryClient, QueryCellarIDsRequest, QueryCellarIDsResponse};
+use tonic::{transport::Channel, Response, Status};
+
+/// Client wrapper used to refresh the cache
+pub struct CorkQueryClient {
```
EricBolten, line 10: Should we also create a `CorkMsgClient` in here for the message sender for consistency?

steward/src/cork/proposals.rs
```diff
@@ -0,0 +1,372 @@
...
+
+use crate::{
+    cellars::{aave_v2_stablecoin, cellar_v1},
+    config::StewardConfig,
+    cork::schedule_cork,
+    error::Error,
+    prelude::APP,
+};
+
+type GovQueryClient = gravity_bridge::gravity_proto::cosmos_sdk_proto::cosmos::gov::v1beta1::query_client::QueryClient<tonic::transport::channel::Channel>;
```
EricBolten: It looks like these two clients are normal Cosmos clients, but we're retrieving them in this indirect and roundabout way because we are already importing `gravity_bridge`. Should we directly import these ourselves? Is this something `deep_space` handles?

steward/src/cork/proposals.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    tokio::spawn(async move {
+        let mut fail_count = 0;
+        loop {
+            tokio::time::sleep(query_period).await;
+
+            debug!("polling for new approved scheduled cork proposals");
+            state
+                .update_latest_observed_height(config.cosmos.grpc.clone())
+                .await;
```
EricBolten, line 62: Is it not also possible for the loop to "fail" during this stage either?

steward/src/cork/cache.rs
```diff
@@ -0,0 +1,69 @@
...
+    tracing::{debug, error},
+    Application,
+};
+use lazy_static::lazy_static;
+use std::{collections::HashSet, sync::RwLock, time::Duration};
+use tokio::task::JoinHandle;
+
+use crate::{cork::client::CorkQueryClient, error::Error, prelude::APP};
+
+pub type ApprovedCellarsCache = RwLock<HashSet<String>>;
```
EricBolten, line 11: Is there a reason this state is implemented as a `RwLock` but we don't do that for the proposal state?

steward/src/cork/proposals.rs
```diff
@@ -0,0 +1,372 @@
...
+        limit: 20,
+        offset: state.proposal_query_offset,
+        ..Default::default()
+    };
+    let mut processed_total: u64 = 0;
+
+    loop {
+        debug!("querying proposals");
+        let proposals = gov_client
+            .proposals(tonic::Request::new(gravity_bridge::gravity_proto::cosmos_sdk_proto::cosmos::gov::v1beta1::QueryProposalsRequest {
```
EricBolten: Similarly here, this roundabout type import gets super ugly. If we have to keep it this way, We should probably have an import statement up top for a large portion of these prefixes.

steward/src/cork/proposals.rs
```diff
@@ -0,0 +1,372 @@
...
+                info!(
+                    "submitted cork for cellar {} triggered by proposal {}",
+                    cellar_id, proposal.proposal_id
+                );
+            }
+        }
+        pagination.offset += processed_total;
+    }
+
+    retry_failed_schedulings(
```
EricBolten: Some thoughts here. The current implementation is retrieving every proposal with pagination, storing failed state about each proposal, and then handing these retries asynchronously. I think a simpler way to handle this whole process would just be this:

1. Add the latest observed proposal ID as part of the state.
2. Rather than querying all proposals, query individual proposal IDs in an incrementing loop. You'll have reached the end when it doesn't exist for your incremented ID.
3. In this method, you won't have to decouple retries -- just do retries inline before moving on to the next proposal ID.

steward/src/error.rs
```diff
@@ -47,10 +47,22 @@ pub enum ErrorKind {
     ProviderError,
     /// Strategy Provider call error
     #[error("SP call error")]
-    SPCallError,
+    SPCall,
```
EricBolten: Should `SPCall`/`GovernanceCall` be `SPCallError` etc. given that these are error types? Same below for `Cache`.

steward/src/utils.rs
```diff
@@ -95,3 +95,7 @@ pub fn sp_call_parse_address(address: String) -> Result<H160, Error> {
         Err(err) => Err(sp_call_error(err.to_string())),
     }
 }
+
+pub fn governance_call_error(message: String) -> Error {
```
EricBolten, line 100: Looks like we did define this function, but didn't use it in the parsing.

EricBolten (commented):
Some feedback.

integration_tests/proposal_test.go
cbrit: this is a field name

proto/cellar_v1.proto
cbrit: let's leave it out

integration_tests/proposal_test.go
cbrit: the latter

steward/src/cellars/cellar_v1.rs
cbrit: Looking it up, I think `as u8` would give the index.

steward/src/commands/cork_proposal.rs
cbrit: It is!

integration_tests/proposal_test.go
EricBolten: Oh, weird. I guess that's the type and not the field name. I thought the serialization for camel casing expected lower camel case though, e.g. `cellarV1`. I'm surprised it works.

steward/src/commands/cork_proposal/aave_v2_stablecoin/sweep.rs
cbrit: Yeah this was the solution I found on stack overflow 😅

steward/src/commands/cork_proposal/cellar_v1/set_fees_distributor.rs
cbrit: Mm I'll double check by running the command but I'm pretty sure the 1 is ignored when retrieving prefixes in deep_space; it definitely is in cosmrs (which ocular uses).

EricBolten: Ah, yeah, I failed to notice that this is a `deep_space` function, if it's handling that bit already then this is fine.

steward/src/cork/proposals.rs
cbrit: iirc this is one of the side effects of the dependency spaghetti we have going on. importing it from another crate wasn't compiling because of duplicate definitions or something.

steward/src/cork/cache.rs
cbrit: Only one thread is accessing the proposal state, while this can be accessed by any number of SP requests

steward/src/cork/proposals.rs
cbrit: Yeah that's much better.

steward/src/error.rs
cbrit: They can be, my thinking was it is a little redundant when referencing it as `ErrorKind::SomethingError` like we usually do but if it were to be imported and used without the full path it would be confusing I guess.

steward/src/cork/cache.rs
EricBolten: Yeah, I mean, the only accessors are here in this file and writing is only occurring in the thread, but I guess it's technically safer to implement the lock in case something down the road decided to directly interact with `APPROVED_CELLARS`. A timing issue between the two is somewhat obviated by the manual refresh call if it's missing.

integration_tests/proposal_test.go
cbrit: I realized the same when you commented that. I can make it camel for clarity, it does look a lot like a type name. Same in the other fields.

steward/src/cork/cache.rs
cbrit: I guess it is only reads that are happening on other threads. I'll leave it as you say, but yeah it's probably unnecessary now that you mention it.

steward/src/commands/start.rs
```diff
@@ -20,6 +29,10 @@ impl Runnable for StartCmd {
         let config = APP.config();
         info!("Starting application");
         abscissa_tokio::run(&APP, async {
+            // currently allows the threads to detach since we aren't capturing the JoinHandle
```
EricBolten, line 32: Given that there is no join call, do these shut down cleanly?

cbrit: Do you mean ends gracefully or ends at all? My understanding is it is killed ungracefully when the process exits.

EricBolten: Yeah I guess the process itself doesn't really have a graceful shutdown as it is. Just trying to reason about how the threads are spawned and detached.

cbrit: One thing we may want is some keep-alive functionality which implies joining these threads so that we can respawn them on error.

EricBolten: Join would only complete after all the threads have exited, no? We're currently effectively blocking on `serve` so in theory you could join all three, but I'm not sure that'd fix the restart issue. Plus, the only way for these threads to exit right now is with a panic, right? I think all errors are handled.

cbrit: Yeah all errors are handled, the validator would just have to restart Steward if they weren't submitting proposal corks due to a panic. I haven't looked at what tokio offers by way of keepalive but I guess it would be it's own detached thread that internally has joined threads?

EricBolten: You'd probably have to structure it as a subtask or something, yeah. The weird thing though is that if something in the thread's processing is instigating a panic, restarting it is unlikely to fix the issue. It almost feels like the more reasonable thing to do is crash the entire process, because then maybe the operator would see that and devs can patch the issue. Failing "quietly" and logging is an option, but might not be noticed. This is also assuming that a crashing process would be noticed and reported though, and not just restarted automatically. Otherwise you basically are in the position of waiting for a human (or some bot process) to notice something it expected to happen didn't happen. For an approved cellar cache issue, that's probably relatively straightforward, but very annoying for a governance prop.

EricBolten: In a more traditional setting this would fire an alert in Sentry or something, but we don't support that sort of thing at the moment and validator setups are inconsistent. Might be worth looking into adding an option for exception handling integration (here and elsewhere).

steward/src/cellars/cellar_v1.rs
cbrit: Well, according to stack overflow you can, but the enum has to be `Copy`, which this one cannot be due to some other unmet trait bounds.

steward/src/cork/client.rs
cbrit: Sending uses the `deep_space::Contact`, I'm not sure adding a wrapper would provide any benefit? When we switch to ocular there will be more consistency I think.

integration_tests/proposal_test.go
cbrit: So, funny thing, it actually *fails to decode* if those fields aren't PascalCase. Something special about them being `oneof` variants?

```bash
2022-12-16T15:04:22.289652Z ERROR steward::cork::proposals: failed to decode GovernanceCall JSON ..... unknown variant `cellarV1`, expected `AaveV2Stablecoin` or `CellarV1` at line 4 column 1
```

EricBolten: Weird...well as long as the format the CLI commands output matches this, I guess it's fine.

steward/src/cork/client.rs
EricBolten: Not crucial, just would make the interface feel cleaner I think. It's something we can come back to later.

steward/src/cellars/aave_v2_stablecoin.rs
EricBolten: This still needs to be addressed.

steward/src/commands/cork_proposal/cellar_v1/trust_position.rs
```diff
@@ -10,12 +10,12 @@ use steward_proto::steward::{
...
 #[derive(Command, Debug, Parser)]
 #[clap(
-    long_about = "DESCRIPTION\n\nCalls trustPosition() on the target cellar contract at the specified block height.\nFor more information see https://github.com/PeggyJV/cellar-contracts/blob/main/src/base/Cellar.sol"
+    long_about = "DESCRIPTION\n\nCalls trustPosition() on the target cellar contract at the specified block height.\nFor more information see https://github.com/PeggyJV/cellar-v1_5/blob/release/src/base/Cellar.sol"
 )]
 pub struct TrustPositionCmd {
     #[clap(short, long)]
     /// Address of the contract (position) to trust
-    position: Address,
+    position: EthereumAddress,
```
EricBolten, line 18: It looks like this import change was made in the Aave calls but only here for CellarV1?

steward/src/cork/proposals.rs
```diff
@@ -81,292 +101,220 @@ impl ProposalThreadState {
...
+            };
 
-            if d.4 as u64 >= config.cork.max_scheduling_retries {
-                warn!(
-                    "proposal {} failed to schedule, will not be retried again.",
-                    d.0
-                );
-                remove_indeces.push(i);
-            }
+            return Err(proposal_processing_error(format!(
```
EricBolten: Is this redundant due to the error message based on the `match` above?

steward/src/cork/proposals.rs
```diff
@@ -81,292 +101,220 @@ impl ProposalThreadState {
...
-            |_: String, _: Vec<u8>, _: u64| -> Result<TxResponse, crate::error::Error> {
-                Ok(TxResponse::default())
-            };
-        block_on(retry_failed_schedulings(
-            &mut state,
-            &test_config,
-            passing_function,
-        ));
+async fn confirm_sheduling(
+    state: &mut ProposalThreadState,
```
EricBolten: Does this need to be a mutable reference?

steward/src/cork/proposals.rs
```diff
@@ -81,292 +101,220 @@ impl ProposalThreadState {
...
-                    for d in state.retry_data.iter() {
-                        if d.0 == proposal.proposal_id {
-                            continue;
-                        }
+        debug!(
+            "proposal {} encoded call: {:?}",
+            proposal.proposal_id, encoded_call
+        );
+
+        if let Err(err) = schedule_cork(&cellar_id, encoded_call.clone(), block_height).await {
```
EricBolten: Wouldn't this bit need to be where attempt-limited retry loop comes in? Otherwise what would happen is a `schedule_cork` call that fails to confirm scheduling will keep retrying until the block height passes (and by virtue of that not scheduling future props if one exists) because we don't increment the proposal ID on a failure. Seems like we'd want to retry the send X number of times based on that config value you added, and then if it failed, move on to the next proposal by incrementing.

docs/api/cellar_v1.md
```diff
@@ -4,25 +4,36 @@
...
+    - [CellarV1.SwapParams](#steward-v3-CellarV1-SwapParams)
+    - [CellarV1.SwapPositions](#steward-v3-CellarV1-SwapPositions)
+    - [CellarV1.UniV2SwapParams](#steward-v3-CellarV1-UniV2SwapParams)
+    - [CellarV1.UniV3SwapParams](#steward-v3-CellarV1-UniV3SwapParams)
+    - [CellarV1Governance](#steward-v3-CellarV1Governance)
+    - [CellarV1Governance.InitiateShutdown](#steward-v3-CellarV1Governance-InitiateShutdown)
+    - [CellarV1Governance.LiftShutdown](#steward-v3-CellarV1Governance-LiftShutdown)
+    - [CellarV1Governance.ResetHighWatermark](#steward-v3-CellarV1Governance-ResetHighWatermark)
+    - [CellarV1Governance.SetFeesDistributor](#steward-v3-CellarV1Governance-SetFeesDistributor)
+    - [CellarV1Governance.SetOwner](#steward-v3-CellarV1Governance-SetOwner)
```
EricBolten: I don't think the markdown docs got regenerated based on the latest updates.

go.sum
```diff
@@ -860,8 +860,10 @@ github.com/pborman/uuid v0.0.0-20170112150404-1b00554d8222/go.mod h1:VyrYX9gd7ir
 github.com/pborman/uuid v1.2.0/go.mod h1:X/NO0urCmaxf9VXbdlT7C2Yzkj2IKimNn4k+gtPdI/k=
 github.com/peggyjv/gravity-bridge/module/v2 v2.0.2 h1:8PiwAPbmeItQu/QbRDzcZUz84ce56NLYmqSSXuLxesY=
 github.com/peggyjv/gravity-bridge/module/v2 v2.0.2/go.mod h1:gKOqaz8xccgMEmWaH53n6uCIvRxqPFgya4aHrMi78Fs=
-github.com/peggyjv/sommelier/v4 v4.0.2 h1:q7rfwjAVtVhh1E7kVlH1iHS6Y+JO8UQD/2aQ43yS8Pk=
-github.com/peggyjv/sommelier/v4 v4.0.2/go.mod h1:0YvTb2SCmN7CIUVzfq+xtF+cu9OmTlyRy6F78BcO0iE=
+github.com/peggyjv/sommelier/v4 v4.0.2-0.20221202205228-cc28bb733ded h1:awq39KRrvJmcEll4xxkkYgStflozwhuRuS1RgzsBZ14=
```
EricBolten: Could probably use a `go mod tidy`.

integration_tests/cork_test.go
EricBolten: Not crucial to fix but worth considering, could make the test flaky.

integration_tests/proposal_test.go
EricBolten: Can we confirm the CLI output gives us this casing?

steward/src/cellars/cellar_v1.rs
EricBolten: Bumping this comment as well.

steward/src/cork/proposals.rs
```diff
@@ -0,0 +1,320 @@
...
+            tokio::time::sleep(query_period).await;
+
+            debug!("polling for new approved scheduled cork proposals");
+            state
+                .update_last_observed_height(config.cosmos.grpc.clone())
+                .await;
+            if let Err(err) = poll_approved_cork_proposals(&mut state).await {
+                fail_count += 1;
+                error!(
+                    "proposal {} failed to process {} time(s). latest error: {}",
```
EricBolten: I'm not sure the fail count works the way one would expect as the poll call contains a loop. As an example, if it exits with an error on a proposal, and then the next call it processes a couple in the loop and then exits at some point with another error, the counter would say 2 for the last processed proposal ID when it had only failed once for that prop. Not sure if it's worth doing anything particularly complicated to maintain this count.

steward/src/cork/proposals.rs
```diff
@@ -0,0 +1,320 @@
...
+                fail_count = 0;
+            }
+        }
+    })
+}
+
+#[derive(Debug, Default)]
+struct ProposalThreadState {
+    last_observed_height: u64,
+    last_processed_proposal_id: u64,
```
EricBolten, line 77: I guess the default value of a u64 is zero then, so we don't need to be explicit for the base case.

EricBolten (commented):
Some follow-up thoughts.

steward/src/cork/proposals.rs
cbrit: Technically no, I just opted not to make another reference that is non-mut for the sake of this function but it's kind of silly in retrospect.

steward/src/cork/proposals.rs
cbrit: I opted to let the polling interval be a "built in" retry, had actually forgot about the config value. I'll add a loop.

EricBolten: Yeah, the thing is that the polling interval as retry doesn't have a stopping point. If you want it to move on, it would be preferable for it to do so after a limited period of time or attempts rather than waiting until the block height it was supposed to execute at before looking past that proposal ID.

steward/src/cork/proposals.rs
```diff
@@ -261,44 +256,51 @@ async fn poll_approved_cork_proposals(state: &mut ProposalThreadState) -> Result
...
-                    return Err(proposal_processing_error(format!(
-                        "failed to confirm cork scheduling for proposal {} reason: {}",
-                        proposal.proposal_id, err
-                    )))
-                }
-            };
+                    Err(err) => {
+                        attempts += 1;
+                        if attempts > config.cork.max_scheduling_retries {
+                            return Err(proposal_processing_error(format!(
```
EricBolten: I think we need to increment the ID here before returning after `max_scheduling_retries` has been reached.

steward/src/cork/proposals.rs
```diff
@@ -261,44 +256,51 @@ async fn poll_approved_cork_proposals(state: &mut ProposalThreadState) -> Result
...
+                                "failed to schedule cork for proposal {} reason: {}",
+                                proposal.proposal_id, err
```
EricBolten: `err` here will be an error returned by `confirm_scheduling` rather than `schedule_cork`, which I think is the error we'd be interested in logging.

steward/src/cork.rs
```diff
@@ -10,74 +13,57 @@ use abscissa_core::{
...
-                warn!("failed to build cork for cellar {}: {}", cellar_id, err);
+                warn!("failed to get encoded call for {}: {}", cellar_id, err);
                 return Err(Status::new(Code::InvalidArgument, err.to_string()));
             }
         };
-        debug!("cork: {:?}", cork);
+        debug!("cork: {:?}", encoded_call);
 
-        if let Err(err) = send_cork(cork).await {
+        if let Err(err) = schedule_cork(&cellar_id, encoded_call.clone(), height).await {
```
EricBolten, line 66: Should we be applying the same retry and error handling logic we've introduced for proposals in our main cork scheduling flow?

EricBolten (commented):
Some final comments.

steward/src/cork/proposals.rs
cbrit: see line 263, i log the schedule_cork error for this reason and then this one is returned as the final failure message. think it should be changed?

steward/src/cork.rs
cbrit: discussed in chat, future PR

steward/src/cork/proposals.rs
EricBolten: I think the only thing is that the log line is indicating the reason for the failure is not seeing the confirmation. This made me realize another issue though: confirm_scheduling returns an `Ok(bool)`. The error match will only happen if the call to confirm the scheduling fails. `Ok(true)` means it was found, but `Ok(false)` means it needs to retry but we aren't handling this condition.

cbrit: I was referring to this line. maybe better as a warning, or we can just only log in each condition branch
```rust
error!("error scheduling cork. confirming if the cork was scheduled in case this is a timeout bug: {}", err);
```

good catch

EricBolten (approved):
LGTM

PR #103, opened 2022-04-08
tweak some config defaults

cbrit:


hannydevelop (approved):
I think amount as 0.00 will suit better though. Not important!

EricBolten (approved):
LGTM

PR #104, opened 2022-04-08
Integration test fixes for sommelier v4

EricBolten:
Integration test fixes for sommelier v4.


integration_tests/cork_test.go
```diff
@@ -94,7 +93,7 @@ func (s *IntegrationTestSuite) TestCork() {
 
 			// make it rain
 			for i := range s.chain.stewards {
-				addr := fmt.Sprintf("localhost:%v", port+i)
+				addr := fmt.Sprintf("localhost:%v", 5734+i)
```
cbrit, line 96: hrm I guess I left this as a magic number everywhere else. Would be good to put it in a var block in setup_tests.go, but it's whatevs.

cbrit (approved):

PR #106, opened 2022-04-11
Collin/config fixes

### LINKED ISSUE #107: Minor improvements
- Default configuration fixes
- Hardcode bech32 prefix in cork handler

### LINKED ISSUE #108: Update Steward Docs

cbrit:
Some fixes to defaults and other tweaks.

The prefix in the config file is no longer used in the cork handler because if changed it could break things. The prefix config field should only be used for other commands, whereas the server will always be using sommelier-prefixed addresses. 

Leveraged the `#[serde(default)]` directive more so we could remove `Option<T>` config fields and the address/port `const`s I put in `server.rs`.

Renamed `rebalancer_key` field to `delegate_key` for clarity

Additionally updated README and added more documentation to the `docs/` directory

Closes #107 
Closes #108


hannydevelop (commented):
You'll need to run fmt.

cbrit:
> You'll need to run fmt.

whoops! done

hannydevelop (approved):

steward/src/config.rs
```diff
@@ -209,7 +209,7 @@ impl Default for GravitySection {
     fn default() -> Self {
         Self {
             contract: "0x0000000000000000000000000000000000000000".to_owned(),
-            fees_denom: "stake".to_owned(),
+            fees_denom: "usomm".to_owned(),
```
cbrit, line 159: @hannydevelop I definitely missed this. Next time would you mind pointing it out in a review instead of committing to my branch?

hannydevelop: yea, no P. I just figured this one commit shouldn't make a PR

cbrit: thanks!

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,75 @@
+# Validators Instructions for Setting Up Steward
+
+Steward is an application intended for developers and validators on the Sommelier network.
+
+It can run as a server for forearding cellar contract calls from Strategy Providers through the Sommelier chain, or it can run in test mode to directly interact with Ethereum contracts as a single signer.
```
EricBolten: forearding -> forwarding

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,75 @@
...
+
+```bash
+steward -c [path to your config toml] start
+```
+
+## Setting Up Steward
+
+In this section, let’s explore setting up steward for validators. There are two ways that validators will use Steward:
+
+1. Running Steward as a gRPC server for relaying SP contract calls through the chain. If a validator is not running this server, it cannot participate in Cellar management.
```
EricBolten: This is the first time in this doc that "SP" is referenced, we should define it.

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,75 @@
...
+2. Manually scheduling contract calls when the validator set needs to coordinate a function  call on a cellar. An example of this would be transferring a Cellar’s accrued fees to a module account on chain.
```
EricBolten: I believe SPs will also be doing this (they don't need on-chain scheduling because they have an external clock they can follow). Scheduled calls are for when multiple parties need to coordinate for valset consensus, like shutting down a cellar.

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Running Steward as a server
+
+Steward runs on every Validator in the Sommelier Validator set. It runs a server to which Strategy Providers (SPs) send requests whenever they determine that the market has changed enough to warrant action. The request payload contains everything needed to make a *cork*: a signed combination of a cellar ID and an ABI encoded contract call. When Steward receives a submission from the SP, it validates the target cellar ID, build a cork, and submits it to the Cork module on chain.
```
EricBolten: Currently there are no "cellar IDs", just target contract addresses.

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,75 @@
...
+
+Here is an ***example*** TOML file with the minimum required configuration fields to run Steward as a server and facilitate Cellar operations. *Please fill in with your own values*:
```
EricBolten: We already have a version of this in the earlier "Config" document. Should we just refer to it there? Possibly copy over the comments too?

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,75 @@
...
+
+```toml
+[cork]
+# Before Steward forwards a function call to the chain, it checks
+# that the target contract address is in fact a cellar approved
+# by governance. To speed up this check, steward frequently queries
+# and caches a list of all approved cellar addresses. This value
+# determines how frequently (in seconds) steward makes this query.
+cache_refresh_period = 60           # default: 60
```
EricBolten: I don't think this code has been merged yet.

docs/03-TheOrchestrator.md
```diff
@@ -0,0 +1,95 @@
+# Orchestrator
+
+The Orchestrator is a sidecar application to a Cosmos chain responsible for *orchestrating* transactions between Cosmos and Ethereum chains. In the Cosmos -> Ethereum direction, validators running the orchestrator with the Relayer enabled will query the [Gravity Module](https://github.com/peggyjv/gravity-bridge/tree/main/module/x/gravity) for pending transactions, and then send them to the gravity contract. In the Ethereum -> Cosmos direction, the Orchestrator acts as an oracle for the Cosmos chain to inform it of incoming transactions. This allows the coordination of appropriate locking/unlocking/minting/burning of tokens on either side of the bridge. The Orchestrator ensures that Gravity module state on the Cosmos chain is synchronized with the state of the gravity contract.
```
EricBolten: We're currently not encouraging validators to run the relayer.

docs/03-TheOrchestrator.md
```diff
@@ -0,0 +1,95 @@
...
+# Omitting the mnemonic argument will provide a prompt for its entry when the
+# command is run.
+steward -c <config toml path> keys cosmos recover <key_name> [mnemonic]
+```
+
+To confirm it works, check your keystore directory for a file with the key name you provided.
+
+### Configuration
+
+The following configuration fields are required to run the Orchestrator. Please refer to the [Configuration Reference](./configuration#reference) for explanations of each fields' purpose.
```
EricBolten: This is the third example config in the documents. It would be good to point to a single exhaustive version.

docs/03-TheOrchestrator.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Quickstart
+
+The Orchestrator and Relayer can be started using the following command:
+
+```bash
+steward -c <config toml path> orchestrator start --ethereum-key <eth_key_name> --cosmos-key <cosmos key name>
+```
+
+If you do not wish to participate in Relaying Cosmos -> Ethereum transactions, you can start the orchestrator without the relayer by adding the `--orchestrator-only` flag:
```
EricBolten: Given the above comment about relaying, we should probably flip the order of these and encourage people to run `--orchestrator-only` as the default for now.

docs/03-TheOrchestrator.md
```diff
@@ -0,0 +1,95 @@
...
+key_derivation_path = "m/44'/60'/0'/0/0"
+
+[keys]
+keystore = "/my/keystore/path"
+```
+
+To add or import an Ethereum key to your keystore, run either of the following commands respectively:
+
+```bash
+steward -c <config toml path> keys eth add <key_name>
```
EricBolten, line 42: This is a more general question, but the difference between `add` and `import` seems unclear on the face of it. IIRC, `add` is generating a new key, so maybe we should call it `generate` or something.

docs/03-TheOrchestrator.md
```diff
@@ -0,0 +1,95 @@
...
+key_derivation_path = "m/44'/118'/0'/0/0"
+
+[keys]
+keystore = "/my/keystore/path"
+```
+
+To add a new Cosmos key or recover one from a mnemonic, run either of the following commands respectively:
+
+```bash
+steward -c <config toml path> keys cosmos add <key_name>]
```
EricBolten: There's a dangling `]` on the end here.

steward/src/allocation.rs
```diff
@@ -231,12 +231,12 @@ pub async fn direct_rebalance(
 ) -> Result<(), Error> {
     let mut tick_info: Vec<CellarTickInfo> = Vec::new();
     let config = APP.config();
-    let keystore = path::Path::new(&config.keystore);
+    let keystore = path::Path::new(&config.keys.keystore);
```
EricBolten: So, we used to have two keystore references: one in `config.keystore` and one in `config.keys.keystore`. It was a conscious choice in a recently merged PR to just use a global keystore config (it's necessary for every function of steward). I think the `keys` stanza in this case would just be a list of specific keys that could be in use.

steward/src/config.rs
```diff
@@ -30,19 +30,18 @@ lazy_static! {
...
 #[serde(default, deny_unknown_fields)]
 pub struct StewardConfig {
-    pub keystore: String,
-    pub server: ServerSection,
-    pub gravity: GravitySection,
-    pub ethereum: EthereumSection,
-    pub cosmos: CosmosSection,
-    pub metrics: MetricsSection,
     pub cellars: Vec<CellarConfig>,
+    pub cosmos: CosmosSection,
```
EricBolten, line 35: I appreciate the idea behind alphabetizing these, but I think having the config divided up by sections grouped by their relevancy is probably more useful to someone reading the config, e.g. global config, an orchestrator/relayer grouping, a steward grouping, etc.

steward/src/config.rs
```diff
@@ -72,48 +71,51 @@ impl StewardConfig {
...
+    pub delegate_key: String,
+    pub keystore: String,
 }
 
 impl Default for KeysConfig {
     fn default() -> Self {
         Self {
-            rebalancer_key: "".to_owned(),
+            delegate_key: "".to_owned(),
+            keystore: "/tmp/keystore".to_owned(),
```
EricBolten: I think guessing a default keystore directory (as has been in the code for some time) is probably unhelpful, and `/tmp` is quite a poor place to store keys as it is not required to persist on system reboots.

steward/src/config.rs
```diff
@@ -175,23 +177,23 @@ impl Default for TokenInfo {
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(default)]
 pub struct EthereumSection {
+    /// Declaring EthereumSection blocks to search
```
EricBolten: We should probably drop these comments, they don't provide additional value.

EricBolten (commented):
Some thoughts.

docs/02-StewardForValidators.md
cbrit: What you explained about the valset is what I was trying to say here. Should I word it more clearly?

EricBolten: It shouldn't require a scheduled cork to transfer accrued fees.

docs/02-StewardForValidators.md
cbrit: The earlier config doc is all available configuration fields, this one is just the required fields for running Steward in "cosmos mode" as we've been calling it. I give "minimum required" config examples throughout and this is meant to be like that. Is it confusing?

docs/03-TheOrchestrator.md
cbrit: agree, I think `new` or `create` are pretty standard

steward/src/allocation.rs
cbrit: Ah, I was unaware of the discussion around it.

steward/src/config.rs
cbrit: I was going to make an issue for refactoring config into steward/orchestrator tables but thought it would be left for later. None of the tables are really clear about what parts of steward used them. Should I change the ordering again or make an issue?

steward/src/config.rs
cbrit: Hah I actually made an issue that cited this behavior as surprising if you forget your config arg. I'll change it to an empty string and make sure it gives a sensible error if you try to use it.

docs/02-StewardForValidators.md
cbrit: Oh I get what you mean now. K

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,36 @@
+# Validators Instructions for Setting Up Steward
+
+Steward is an application intended for developers and validators on the Sommelier network.
```
elsehow: An application... that does what? Linking to a doc that explains what Steward does at a high level, and how it fits into the broader ecosystem of things Sommelier validators need to run or worry about, could be helpful here.

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,36 @@
...
+
+It can run as a server for forwarding cellar contract calls from Strategy Providers through the Sommelier chain, or it can run in test mode to directly interact with Ethereum contracts as a single signer.
+
+It integrates the full functionality of gorc for operating as an orchestrator and relayer of [Gravity bridge](https://github.com/PeggyJV/gravity-bridge/) messages between the Ethereum and Cosmos chains.
+
+Stewards works in conjunction with the Orchestrator, so both processes must be running to fully participate in Cellar management.
```
elsehow: Linking to docs on the Orchestrator would be good here as well.

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,36 @@
...
+## Setting Up Steward
+
+In this section, let’s explore setting up steward for validators. There are two ways that validators will use Steward:
+
+1. Running Steward as a gRPC server for relaying Strategy Provider contract calls through the chain. If a validator is not running this server, it cannot participate in Cellar management.
+2. Manually scheduling contract calls when the validator set needs to coordinate administrative function calls on a cellar. An example of this would be shutting down a cellar.
+
+### Running Steward as a server
+
+Steward runs on every Validator in the Sommelier Validator set. It runs a server to which Strategy Providers (SPs) send requests whenever they determine that the market has changed enough to warrant action. The request payload contains everything needed to make a *cork*: a signed combination of a cellar address and an ABI encoded contract call. When Steward receives a submission from the SP, it validates the target cellar address, builds a cork, signs it with the delegate key, and submits it to the Cork module on chain.
```
elsehow: This is an extremely helpful overview of what Steward does. Breaking this information out and putting it in some overview document that other docs can link to would be great. (See comment above).

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,36 @@
...
+
+> :bulb:  Steward gatekeeps function calls both by only supporting certain Cellar functions in its API, and by validating the target cellar before forwarding a call.  
+
+Before SPs can establish a connection with your `steward` server, you will need to generate a self-signed CA certificate, a certificate signed by this CA, and you'll need to the SP's client CA. These certificates must be TLS 1.2 compliant. The paths to these values and the key used to create your signed server certificate must be set in the `[server]` table of your config file.
```
elsehow: I think you mean "TLS certificate." A link to how to generated a self-signed TLS certificate could be helpful.

docs/03-TheOrchestrator.md
```diff
@@ -0,0 +1,61 @@
+# Orchestrator
+
+The Orchestrator is a sidecar application to a Cosmos chain responsible for *orchestrating* transactions between Cosmos and Ethereum chains. In the Cosmos -> Ethereum direction, validators running the orchestrator with the Relayer enabled will query the [Gravity Module](https://github.com/peggyjv/gravity-bridge/tree/main/module/x/gravity) for pending transactions, and then send them to the gravity contract. In the Ethereum -> Cosmos direction, the Orchestrator acts as an oracle for the Cosmos chain to inform it of incoming transactions. This allows the coordination of appropriate locking/unlocking/minting/burning of tokens on either side of the bridge. The Orchestrator ensures that Gravity module state on the Cosmos chain is synchronized with the state of the gravity contract. Currently we are only asking validators to run the Orchestrator without the Relayer (Ethereum -> Cosmos direction only).
```
elsehow: Great summary - something like this would address my comments on Steward (above).

docs/02-StewardForValidators.md
cbrit: Have another PR with a script/doc for generating the CA and self-signed TLS cert. Will update the docs with a link when it's merged!

README.md
```diff
@@ -1,86 +1,36 @@
 # Steward
 
-Steward is an application intended for developers and validators on Sommelier network.
-
-It can run as a voter in the Cosmos Sommelier protocol or in test mode to directly interact with Ethereum contracts as a single signer.
-
-It integrates the full functionality of [Gorc](https://github.com/PeggyJV/gravity-bridge/tree/main/orchestrator/gorc) for operating as an orchestrator and relayer of [Gravity bridge](https://github.com/PeggyJV/gravity-bridge) messages between the Ethereum and Cosmos chains.
+Steward is a key piece of the Sommelier system. It is a side-car process that connects the Sommelier chain to the outside world. It provides an API for Strategy Providers to rebalance Cellars through the Sommelier validator set based on off-chain calculations. Additionally, it subsumes the functionality of gorc for managing the [Gravity bridge Orchestrator](https://github.com/PeggyJV/gravity-bridge/tree/main/orchestrator), and provides developer tools and features for developing and testing Cellars.
```
EricBolten: "rebalance" is a pretty specific term, "update" might be more appropriate here.

README.md
```diff
@@ -1,86 +1,36 @@
...
-```
-cargo run --bin steward -- -c [your_config_file_name.toml] keys eth import [key_name]
-```
+Steward wears several hats:
 
-Now, navigate to the keystore location in your local environment, i.e `keys.keystore`. Confirm that the key was created successfully in the location you specified in your config file.
+###  1. Gravity Bridge Operator
 
-#### Authorize Erc20 Token to interact with Cellar contract.
+Steward is responsible for running the Relayer, which handles relaying Cosmos transactions to Ethereum, and the Orchestrator, which enables the co-processing of Ethereum transactions on Sommelier. Steward runs the Orchestrator so that Sommelier can manage Cellars on Ethereum. Which brings us to Steward's second responsibility.
```
EricBolten: The last sentence here ("Which brings us...") is unnecessary.

README.md
```diff
@@ -1,86 +1,36 @@
...
 
-Run the `allow-erc-2-0` command as shown below, to allow Erc20 token to interact with Cellar contract.
+### 2. Middleman between the Sommelier chain and Strategy Providers
 
-```
-cargo run --bin steward -- -c [your_config_file_name.toml] allow-erc-2-0 --cellar-address=[the_cellar_address] address [the_erc20_address] --amount [amount]
+Steward runs on every Validator in the Sommelier Validator set. It runs a server to which Strategy Providers (SPs) send requests whenever they determine that the market has changed enough to warrant action. The request payload contains everything needed to make a *cork*: a signed combination of a cellar address and an ABI encoded contract call. When Steward receives a submission from the SP, it validates the target cellar address, builds a cork, signs it with the delegate key, and submits it to the [Cork module](https://github.com/PeggyJV/sommelier/tree/main/x/cork) on chain, where it will be evaluated before being bridged to the target Cellar contract.
```
EricBolten: "warrant action" is very vague here, should probably indicate this is talking about making a call to a cellar to update it.

EricBolten: In the last sentence, "where it will be evaluated...", might be more appropriate to say where it will be bridged to the target cellar contract if a consensus of validators submit the same cork.

README.md
```diff
@@ -1,86 +1,36 @@
...
-Next, create a `toml` file in the root of the application, replacing the default keys in the template displayed in your terminal with your configuration. Make sure to confirm that the token info in your configuration file matches the deployed Cellar contract.
-
-You can create keys or import keys. To create keys, run the command below. Replace `eth` with `cosmos` if you want to create a `cosmos` key instead of an `eth` key.
-
-```
-cargo run --bin steward -- -c [your_config_file_name.toml] keys eth add [key_name]
-```
-
-If you already have a key, you can import it with the command below:
+## Steward use case
```
EricBolten: case -> cases

README.md
```diff
@@ -1,86 +1,36 @@
...
 
-```
-cargo run --bin steward -- -c [your_config_file_name.toml] keys eth import [key_name]
-```
+Steward wears several hats:
```
EricBolten: Unnecessary line, delete it.

README.md
```diff
@@ -104,4 +54,4 @@ Below is a list of the Steward's subcommands:
 | tx                | Create transactions on either ethereum or cosmos chains    |
 | allow-erc-2-0     | Allow Erc20 Token to interact with cellar contract         |
 
-If you have any questions, you can ask the community in our [Telegram](https://t.me/getsomm) and [Discord](https://discord.com/invite/ZcAYgSBxvY) group. Also, you can take a look at the [documentation](docs/validators_instructions) for Validators instructions.
+If you have any questions, you can ask the community in our [Telegram](https://t.me/getsomm) and [Discord](https://discord.com/invite/ZcAYgSBxvY) group.
```
EricBolten, line 52: In the section above, we've altered and removed some of these commands. Let's get this table up to date.

steward/src/commands/fund_cellar.rs
```diff
@@ -0,0 +1,135 @@
+#![allow(unused_variables)]
```
EricBolten: Your branch is re-adding these deleted commands, will have to remove this file and some others that were added back here.

steward/src/config.rs
```diff
@@ -14,10 +14,10 @@ use signatory::FsKeyStore;
 use std::{net::SocketAddr, path::Path, time::Duration};
 
 lazy_static! {
-    pub static ref DELEGATE_KEY: PrivateKey = {
+    pub static ref DELEGATE_KEY: cosmos_gravity::crypto::PrivateKey = {
```
EricBolten: There's no reason internally to swap to this key type. It's only necessary for gravity bridge functions that expect this type, and it's specific to feature flags for ethermint. We should use the regular `deep_space` `PrivateKey` wherever possible inside Steward.

steward/src/config.rs
EricBolten: I think it's fine to leave it as you have it here for now.

steward/src/somm_send.rs
```diff
@@ -1,9 +1,9 @@
 use deep_space::coin::Coin;
 use deep_space::error::CosmosGrpcError;
-use deep_space::private_key::PrivateKey as CosmosPrivateKey;
 use deep_space::Contact;
 use deep_space::Fee;
 use deep_space::Msg;
+use gravity_bridge::cosmos_gravity::crypto::PrivateKey as CosmosPrivateKey;
```
EricBolten: Same thing here, let's not use this type unless a gravity bridge function specifically requires it.

docs/01-Configuration.md
```diff
@@ -0,0 +1,322 @@
...
+### Top-level table
+
+#### `keystore`
+
+Type: string
+
+Path to the location where `steward keys` will manage keys on-disk
+
+```
+keystore = "/tmp/keystore"
```
EricBolten: Let's not suggest that people put their keystore in `/tmp`.

docs/01-Configuration.md
```diff
@@ -0,0 +1,322 @@
...
+To generate a default config file, simply redirect stdout to a file:
+
+```bash
+# You must update these values to your own configuration
+steward print-config > config.toml
+```
+
+## Reference
+
+### Top-level table
```
EricBolten, line 22: I'm confused by the use of the word "table" in this doc. In TOML there are only config key-value pairs and section names.

docs/01-Configuration.md
```diff
@@ -0,0 +1,322 @@
...
+
+### `[cosmos.gas_price]` table
+
+For setting transaction fees
+
+#### `amount`
+
+Type: float
+
+The amount of `denom` you are willing to pay per gas for transactions.
```
EricBolten, line 101: We should explicitly call out that we expect people to set this to 0.0 for now.

docs/01-Configuration.md
```diff
@@ -0,0 +1,322 @@
...
+
+### `[keys]` table
+
+Keys config
+
+#### `delegate_key`
+
+Type: string
+
+The name of the key in `keystore` used for delegate signing by both Steward and the Orchestrator. The key name is the name of the key file *without the file extension*.
```
EricBolten, line 203: We should explain somewhere to validators that they should be using the same key material as their orchestrator key which has already been registered on-chain. Additionally, if Steward is subsuming the functions of `gorc`, there needs to be a method for new validators to get their delegate key signatures and register their keys on-chain. I think at the moment you'd still have to use `gorc` to do that.

docs/01-Configuration.md
```diff
@@ -0,0 +1,322 @@
...
+```
+[server]
+address = "0.0.0.0"
+```
+
+#### `client_ca_cert_path`
+
+Type: string
+
+The path to the trusted CA used to sign the Strategy Provider's client certificate
```
EricBolten, line 244: If fields are left out of the config, they are set to the default, correct? No one should be setting the client CA, we want to default to the current hardcoded PeggyJV CA. If there are other defaults we don't want people messing with, we should probably ask them not to set them.

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,31 @@
+# Validators Instructions for Setting Up Steward
+
+Steward is a side-car process that connects the Sommelier chain to the outside world. It provides an API for Strategy Providers to rebalance Cellars through the Sommelier validator set based on off-chain calculations. Additionally, it subsumes the functionality of gorc for managing the [Gravity bridge Orchestrator](https://github.com/PeggyJV/gravity-bridge/tree/main/orchestrator), and provides developer tools and features for developing and testing Cellars.
```
EricBolten: This section is a copy-paste of the top-level README intro. Do we want to duplicate this? If we do, we should also make sure that they are both updated in tandem.

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,31 @@
...
+
+Steward runs on every Validator in the Sommelier Validator set. It runs a server to which Strategy Providers (SPs) send requests whenever they determine that the market has changed enough to warrant action. The request payload contains everything needed to make a *cork*: a signed combination of a cellar address and an ABI encoded contract call. When Steward receives a submission from the SP, it validates the target cellar address, builds a cork, signs it with the delegate key, and submits it to the [Cork module](https://github.com/peggyjv/sommelier/tree/main/x/cork) on chain.
+
+> :bulb:  Steward gatekeeps function calls both by only supporting certain Cellar functions in its API, and by validating the target cellar before forwarding a call.
```
EricBolten: Suggestion: "gatekeeps" -> "authorizes".

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,31 @@
...
+Steward runs on every Validator in the Sommelier Validator set. It runs a server to which Strategy Providers (SPs) send requests whenever they determine that the market has changed enough to warrant action. The request payload contains everything needed to make a *cork*: a signed combination of a cellar address and an ABI encoded contract call. When Steward receives a submission from the SP, it validates the target cellar address, builds a cork, signs it with the delegate key, and submits it to the [Cork module](https://github.com/peggyjv/sommelier/tree/main/x/cork) on chain.
+
+> :bulb:  Steward gatekeeps function calls both by only supporting certain Cellar functions in its API, and by validating the target cellar before forwarding a call.  
+
+## Quickstart
+
+To start Steward, simply run
+
+```bash
+steward -c [path to your config toml] start
```
EricBolten: Nit: use `<blah>` for required options rather than `[blah]` which indicates optional ones.

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,31 @@
...
+```
+
+> :warning: If this is production, don't forget to make sure the Orchestrator is also running! If you don't, you can be jailed and your submitted corks will not be processed.
+See the [Orchestrator Quickstart](./docs/03-TheOrchestrator.md#quickstart) section of the docs. 
+
+## Setting Up Steward
+
+In this section, let’s explore setting up steward for validators.
```
EricBolten: Unnecessary sentence, drop it.

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Running Steward as a server
+
+Before SPs can establish a connection with your `steward` server, you will need to generate a CA, a TLS certificate signed by this CA, and you'll need to the SP's client CA. These certificates must be TLS 1.2 compliant. The paths to these values and the key used to create your signed server certificate must be configured in the `[server]` table of your config file. Starting out, Sommelier will be running the only SP, and our client CA cert can be found in the `tls/` directory of this repo. You will use this path for the `client_ca_cert_path` field in your config TOML.
```
EricBolten: TLS 1.2 -> TLS 1.3

EricBolten: This is reiterating the earlier point -- don't we hardcode the default cert into the binary? Why would we require people to point to it in their config? It's not expected that an operator will have checked out the repo, our release process will build a steward binary that can be downloaded independently, and shouldn't rely on external files to run.

docs/03-TheOrchestrator.md
```diff
@@ -0,0 +1,61 @@
...
+
+```bash
+steward -c <config toml path> orchestrator start --orchestrator-only --ethereum-key <eth_key_name> --cosmos-key <cosmos key name>
+```
+
+The `--orchestrator-only` flag prevents the Relayer thread from running, and this is the way in which we encourage validators to run the Orchestrator for now. You will burn gas on failed transactions if you run the Relayer in its current state.  
+
+## Setup
+
+Before you'll be able to run the start command successfully however, you'll need keys, a configuration file, and a running Cosmos chain. Let's walk through these one at a time.
```
EricBolten: "successfully however," -> "successfully"

EricBolten: You can drop "Let's walk through...", it's unnecessary.

docs/03-TheOrchestrator.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Key creation config
```
EricBolten, line 23: If this is a new validator, this will generate keys for them, but they will be unable to use them without further instructions as to how to register their delegate keys. If it's an existing validator, we should instruct them to copy their key material into a keystore directory for steward.

EricBolten (commented):
Another round of comments.

docs/01-Configuration.md
cbrit: TOML docs refer to hash tables/dictionaries as tables. https://toml.io/en/v1.0.0#table

I could say section instead?

EricBolten: Weird, guess I was wrong on this one. Their spec would be the right way.

docs/02-StewardForValidators.md
cbrit: I figured Validators would come to this doc more frequently than the README so it might be good to have it in both places. I don't see it changing much but it can go either way for me.

README.md
```diff
@@ -1,107 +1,54 @@
...
 | eth-to-cosmos     | Command to Send Ethereum to Cosmos                         |
+| help              | Help command to get usage information                      |
+| keys              | Key management commands for the rebalancer                 |
 | orchestrator      | The orchestrator management commands                       |
-| query             | Command to query state on either ethereum or cosmos chains |
+| print-config      | Command for printing default configurations                |
 | sign-delegate-key | This command is to sign delegate keys                      |
-| tx                | Create transactions on either ethereum or cosmos chains    |
-| allow-erc-2-0     | Allow Erc20 Token to interact with cellar contract         |
+| transfer          | Command to transfer ETH                                    |
```
EricBolten: Does this actually do what it says? Do we need this command at all?

README.md
```diff
@@ -1,107 +1,54 @@
...
 | cosmos-to-eth     | This command sends Cosmos to the Eth chain                 |
 | deploy            | Provides tools for contract deployment                     |
 | eth-to-cosmos     | Command to Send Ethereum to Cosmos                         |
+| help              | Help command to get usage information                      |
+| keys              | Key management commands for the rebalancer                 |
 | orchestrator      | The orchestrator management commands                       |
-| query             | Command to query state on either ethereum or cosmos chains |
+| print-config      | Command for printing default configurations                |
 | sign-delegate-key | This command is to sign delegate keys                      |
-| tx                | Create transactions on either ethereum or cosmos chains    |
```
EricBolten, line 104: These commands still exist on main, we should probably delete them too.

docs/01-Configuration.md
```diff
@@ -0,0 +1,327 @@
...
+
+[ethereum]
+blocks_to_search = 5000
+gas_price_multiplier = 1.0
+key_derivation_path = "m/44'/60'/0'/0/0"
+rpc = "http://localhost:8545"
+
+[gravity]
+contract = "0x0000000000000000000000000000000000000000"
+fees_denom = "stake"
```
EricBolten: "stake" -> "usomm"

docs/01-Configuration.md
```diff
@@ -0,0 +1,327 @@
...
+
+[keys]
+delegate_key = ""
```
EricBolten: Since this is an example config, let's put an example name in here, e.g. "key_name" or something.

docs/01-Configuration.md
```diff
@@ -0,0 +1,327 @@
...
+
+[metrics]
+listen_addr = "127.0.0.1:3000"
+
+# Please leave the client_ca_cert_path field unset for now
+[server]
+address = "0.0.0.0"
+port = 5734
+server_cert_path = ""
```
EricBolten: Similar to keystore, let's indicate "/server/cert/path" rather than an empty value.

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,40 @@
+# Validators Instructions for Setting Up Steward
+
+Steward is a side-car process that connects the Sommelier chain to the outside world. It provides an API for Strategy Providers to rebalance Cellars through the Sommelier validator set based on off-chain calculations. Additionally, it subsumes the functionality of [gorc](https://github.com/PeggyJV/gravity-bridge/tree/main/orchestrator/gorc) for managing the [Gravity bridge Orchestrator](https://github.com/PeggyJV/gravity-bridge/tree/main/orchestrator), and provides developer tools and features for developing and testing Cellars.
+
+Steward runs on every Validator in the Sommelier Validator set. It runs a server to which Strategy Providers (SPs) send requests whenever they determine that the market has changed enough to warrant action. The request payload contains everything needed to make a *cork*: a signed combination of a cellar address and an ABI encoded contract call. When Steward receives a submission from the SP, it validates the target cellar address, builds a cork, signs it with the delegate key, and submits it to the [Cork module](https://github.com/peggyjv/sommelier/tree/main/x/cork) on chain.
```
EricBolten: Same comment here re: "warrant action" vs. "update a cellar". The perils of copy-pasting sections that get updated!

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+> :warning: If this is production and your node is running, don't forget to make sure the Orchestrator is also running! If you don't, you can be jailed and your submitted corks will not be processed.
+See the [Orchestrator Quickstart](./docs/03-TheOrchestrator.md#quickstart) section of the docs.
+
+## Running Steward as a server
+
+### TLS Certificates
+
+Before SPs can establish a connection with your `steward` server, you will need to generate a CA, a TLS certificate signed by this CA, and you'll need to the SP's client CA. These certificates must be TLS 1.3 compliant. The paths to these values and the key used to create your signed server certificate must be configured in the `[server]` table of your config file. Starting out, Sommelier will be running the only SP, and Steward will use our client CA cert by default. This CA cert can be found in the `tls/` directory of this repo. There is no need to set the `client_ca_cert_path` configuration field at this time.
```
EricBolten: "and you'll need to the SP's client CA" needs to be updated to whatever was meant to be said here.

EricBolten: PeggyJV is running the SP, Sommelier is a protocol and not an organization.

EricBolten: Might be worth mentioning specifically that it doesn't need to be set because it will default to the PeggyJV cert included in the binary in this first version.

docs/03-TheOrchestrator.md
```diff
@@ -0,0 +1,61 @@
...
+
+The Orchestrator can be started using the following command:
+
+```bash
+steward -c <config_toml_path> orchestrator start --orchestrator-only --ethereum-key <eth_key_name> --cosmos-key <cosmos_key_name>
+```
+
+The `--orchestrator-only` flag prevents the Relayer thread from running, and this is the way in which we encourage validators to run the Orchestrator for now. You will burn gas on failed transactions if you run the Relayer in its current state.
+
+## Setup
```
EricBolten, line 17: These key generation instructions are the same sort of setup described in the "Steward for Validators" doc. We should also indicate here that they must be registered delegate keys and that old keys from gorc setups can be copied over if they are already registered.

steward/src/commands.rs
```diff
@@ -46,9 +44,6 @@ pub enum StewardCmd {
     #[clap(subcommand)]
     Orchestrator(orchestrator::OrchestratorCmd),
     SignDelegateKeys(sign_delegate_keys::SignDelegateKeysCmd),
-    #[clap(subcommand)]
-    Tx(tx::TxCmd),
-    AllowErc20(allow_erc20::AllowERC20),
```
EricBolten, line 51: If we're deleting these, we should delete the commands themselves and update the main README to not reference them.

steward/src/commands/fund_cellar.rs
EricBolten: Bumping this, these files still need to be removed.

steward/src/commands/print_config.rs
```diff
@@ -1,30 +0,0 @@
-use crate::config::StewardConfig;
-use crate::{application::APP, prelude::*};
-use abscissa_core::{Command, Clap, Runnable};
-
-#[derive(Command, Debug, Default, Clap)]
-pub struct PrintConfigCmd {
```
EricBolten, line 6: I'm pretty sure we're still using and referencing `print-config` so we probably don't want to delete this file.

steward/src/commands/allow_erc20.rs
```diff
@@ -36,7 +36,7 @@ impl Runnable for AllowERC20 {
 
         let name = &config
             .keys
-            .rebalancer_key
+            .delegate_key
```
EricBolten: Tagging this file as meant to be deleted based on `AllowErc20` command being removed.

steward/src/commands/reinvest.rs
```diff
@@ -0,0 +1,66 @@
...
+
+use ethers::prelude::*;
+
+use signatory::FsKeyStore;
+
+use crate::{allocation, cellars::uniswapv3::UniswapV3CellarState, gas::CellarGas, prelude::*};
+
+/// Cellars reinvest command
+#[derive(Command, Debug, Parser)]
+pub struct ReinvestCommand {
```
EricBolten: Delete this file.

steward/src/commands/remove_funds.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use crate::{allocation, cellars::uniswapv3::UniswapV3CellarState, prelude::*};
+use steward_abi::cellar_uniswap::*;
+
+/// Remove funds from Cellars
+#[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Remove funds from Cellar.\n This command remove funds from the required Cellar by loading the Cellar config."
+)]
+pub struct RemoveFundsCmd {
```
EricBolten: Delete this file.

steward/src/commands/set_validator.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use abscissa_core::{clap::Parser, Command, Runnable};
+use ethers::prelude::*;
+use signatory::FsKeyStore;
+
+use crate::{allocation, cellars::uniswapv3::UniswapV3CellarState, gas::CellarGas, prelude::*};
+
+/// Set-validator subcommand
+#[derive(Command, Debug, Parser)]
+pub struct SetValidatorCmd {
```
EricBolten: Delete this file.

EricBolten (commented):
Some further comments and suggestions.

steward/src/commands/print_config.rs
cbrit: `print-config` is using `config-cmd.rs`, maybe a mistake but that's how it's been working.

steward/src/commands/reinvest.rs
cbrit: I deleted these I dont know what happened. here we go again

docs/02-StewardForValidators.md
```diff
@@ -0,0 +1,41 @@
+# Validators Instructions for Setting Up Steward
+
+Steward is a side-car process that connects the Sommelier chain to the outside world. It provides an API for Strategy Providers to rebalance Cellars through the Sommelier validator set based on off-chain calculations. Additionally, it subsumes the functionality of [gorc](https://github.com/PeggyJV/gravity-bridge/tree/main/orchestrator/gorc) for managing the [Gravity bridge Orchestrator](https://github.com/PeggyJV/gravity-bridge/tree/main/orchestrator), and provides developer tools and features for developing and testing Cellars.
```
EricBolten: "rebalance" -> "update"

docs/03-TheOrchestrator.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Quickstart
+
+The Orchestrator can be started using the following command:
+
+```bash
+steward -c <config_toml_path> orchestrator start --orchestrator-only --ethereum-key <eth_key_name> --cosmos-key <cosmos_key_name>
+```
+
+The `--orchestrator-only` flag prevents the Relayer thread from running, and this is the way in which we encourage validators to run the Orchestrator for now. You will burn gas on failed transactions if you run the Relayer in its current state.
```
EricBolten, line 13: Perhaps a brief mention that the `--ethereum-key` and `--cosmos-key` args must be the same keys your validator has registered as delegate keys.

EricBolten (commented):
Two final, minor comments.

EricBolten (approved):
LGTM

PR #117, opened 2022-04-18
Collin/info logs

### LINKED ISSUE #112: Steward INFO logging
Some `debug!` logs need to be changed to `info!`, currently no information is communicated to the operator when steward receives/processes/sends corks.

cbrit:
Closes #112


steward/src/cork.rs
```diff
@@ -87,7 +90,7 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
                 "failed to send cork to sommelier",
             ));
         }
-        debug!("sent cork!");
+        info!("submitted cork for {}!", request.cellar_id);
```
EricBolten: Could we decorate this log message further by including e.g. the type URL of the request, so we know what sort of message is being sent to the cellar? Same would apply to the message at the start of processing as well.

EricBolten (commented):
A suggestion.

steward/src/cork.rs
cbrit: That's a good idea. I was trying to figure out how to log what function was being called and I forgot about the type url.

cbrit: It looks like there is no proto type_url available unless it's a `prost::Any`. There is a Rust type_id which may accomplish the same thing.

steward/src/cellars.rs
```diff
@@ -11,6 +12,10 @@ pub fn validate_cellar_id(cellar_id: &str) -> Result<(), String> {
     Ok(())
 }
 
+pub fn log_cellar_call(cellar_name: &str, function_name: &str, cellar_id: String) {
```
EricBolten: Nit: a tad weird to use multiple different string types here.

steward/src/cellars/aave_v2_stablecoin.rs
```diff
@@ -2,37 +2,47 @@ use ethers::{abi::AbiEncode, prelude::H160};
...
 
-pub fn get_encoded_call(function: Function) -> Vec<u8> {
+use super::log_cellar_call;
+
+const CELLAR_NAME: &str = "aave_v2_stablecoin";
+
+pub fn get_encoded_call(function: Function, cellar_id: String) -> Vec<u8> {
     match function {
         EnterStrategy(_) => {
+            log_cellar_call(CELLAR_NAME, "enterStrategy", cellar_id);
```
EricBolten: Can you get the raw string (or similar) `enterStrategy` by inspecting the `EnterStrategyCall` struct you create here, so that you don't have to hardcode the string?

EricBolten (commented):
Minor comments, though this one probably also should be updated post-merge of the AAVE update.

EricBolten:
Now that the new Aave contract is merged it'd be a good time to update this.

EricBolten (approved):
LGTM

PR #118, opened 2022-04-19
Update `gravity_bridge`

EricBolten:
Update the `gravity_bridge` dependency to the latest commit and fix up references to the wrapped `deep_space` Cosmos key type it now expects.


cbrit (approved):
LGTM

Issue #111, opened 2022-04-14
Environment variable(s) for Steward config


cbrit:
Rather than requiring users to pass their config toml as an argument to every single steward command, it would be good for there to be a STEWARD_CONFIG_PATH environment variable that is checked on startup and used if present. This also prevents the strange behavior of using unexpected default config values when you forget to use the `-c` parameter, such as when creating a key and it appearing in /tmp/keystore by mistake.

Not sure if [Abscissa](https://github.com/iqlusioninc/abscissa) actually supports this currently, it may require a PR there.

hannydevelop:
I think I've seen something like this in clap, not very sure. If clap has it, Abscissa definitely will. So, I'm picking this up since I have the better experience here.

PR #119, opened 2022-04-19
Environment variable(s) for Steward config

hannydevelop:
This PR closes out #111


steward/src/commands.rs
```diff
@@ -85,13 +85,22 @@ impl Configurable<StewardConfig> for EntryPoint {
             .map(PathBuf::from)
             .unwrap_or_else(|| CONFIG_FILE.into());
 
-        if filename.exists() {
+        let config_env_variable = option_env!("STW_ENV_VAR");
```
EricBolten: We don't need to say "ENV_VAR", it's implicit. `STEWARD_CONFIG` would be preferable.

EricBolten: Additionally, `option_env!` is a compile-time macro. We can check for the presence of runtime env vars using `std::env::var("VARIABLE").is_ok()`.

steward/src/commands.rs
```diff
@@ -85,13 +85,22 @@ impl Configurable<StewardConfig> for EntryPoint {
...
+
+        if config_env_variable != None {
+            let new_filename = self.config.as_ref().map(PathBuf::from).unwrap_or_else(|| {
+                config_env_variable
+                    .expect("Can't process config_env_variable")
+                    .into()
+            });
+            println!("{}", new_filename.display());
```
EricBolten: I assume this was here for debugging purposes but we shouldn't print out the filename here.

steward/src/commands.rs
```diff
@@ -85,13 +85,22 @@ impl Configurable<StewardConfig> for EntryPoint {
...
+            Some(new_filename)
```
EricBolten: Let's add a test in our acceptance testing to make sure commands use the correct hierarchy of preference here. Testing with both a command-line config set, both command-line and the env var set, and then with only the env var set.

EricBolten (commented):
Some suggestions.

steward/src/commands.rs
hannydevelop: done

steward/src/commands.rs
hannydevelop: done

steward/src/commands.rs
```diff
@@ -78,13 +78,21 @@ impl Configurable<StewardConfig> for EntryPoint {
             .map(PathBuf::from)
             .unwrap_or_else(|| CONFIG_FILE.into());
 
-        if filename.exists() {
+        let config_env_variable = std::env::var("VARIABLE").is_ok();
```
EricBolten: We shouldn't use `VARIABLE` as our env var name. Let's go with `STEWARD_CONFIG`.

Also, calling `is_ok()` is telling you if the `Result` from `std::env::var()` is `Ok`. In order to retrieve the variable, we'd need to assign `std::env::var("STEWARD_CONFIG")` to a variable, check if it `is_ok()`, and then unwrap it to get the value stored in the env variable.

steward/src/commands.rs
```diff
@@ -78,13 +78,21 @@ impl Configurable<StewardConfig> for EntryPoint {
...
+
+        if config_env_variable {
```
EricBolten: We'll have to update the logic of our overall check here. Before we get to this `if` statement, we should be recording the configured `filename` and the `config_env_variable`. The current code checks for a defined path, and if it doesn't find one, it returns `CONFIG_FILE` as the file path.

Our expected logic is:
1. use the config path if it's provided
2. use the environment variable if it's provided
3. try the default CONFIG path
4. return None

This means that we can't `unwrap_or_else(|| CONFIG_FILE.into()` any more, because that makes cases 1 and 3 happen at the same time, so we'd have to leave it as an `Option` until the `if` check.

Then, we'd have to check
1. `filename.is_some()` and if so, use it
2. `config_env_variable.is_ok()` and if so, use it
3. construct a PathBuf from `CONFIG_FILE` and if it exists, use it
4. return `None`

steward/src/commands.rs
```diff
@@ -78,13 +78,21 @@ impl Configurable<StewardConfig> for EntryPoint {
...
+            let new_filename = self
+                .config
+                .as_ref()
+                .map(PathBuf::from)
+                .unwrap_or_else(|| config_env_variable.to_string().into());
```
EricBolten: In this case we shouldn't try to read `self.config` again, if we're going to use the environment variable we should just construct a PathBuf for it directly.

EricBolten (commented):
Some further comments.

steward/mew.toml
```diff
@@ -0,0 +1 @@
+keystore = "mew"
```
EricBolten: Seems like maybe this was checked in by accident?

steward/mew/hanny.pem
```diff
@@ -0,0 +1,5 @@
+-----BEGIN PRIVATE KEY-----
```
EricBolten: Same here, unintended checkin?

steward/stew.toml
```diff
@@ -0,0 +1 @@
+keystore = "stew"
```
EricBolten: Same here?

steward/tests/acceptance.rs
```diff
@@ -619,3 +619,118 @@ fn print_config() {
...
+    let config_string = toml::to_string(&config).expect("could not write config to TOML string");
+    fs::write(config_file_path.clone(), config_string).expect("could not write config file");
+
+    let env_config_file_path = key_dir.path().join("config.toml");
+    let env_config_string =
+        toml::to_string(&env_config).expect("could not write config to TOML string");
+    fs::write(env_config_file_path.clone(), env_config_string)
+        .expect("could not write config file");
+
+    let config_file_path = key_dir.path().join("config.toml");
```
EricBolten, line 668: These three lines are duplicates of three earlier lines.

EricBolten (commented):
A few comments.

steward/mew.toml
hannydevelop: I was using those for tests. Removing them now

EricBolten: This file still needs to be deleted.

EricBolten (commented):
One last fix.

EricBolten (approved):
LGTM

PR #121, opened 2022-04-20
Steward test mode

hannydevelop:


steward/src/cork.rs
```diff
@@ -94,6 +105,140 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
...
+    async fn submit(
+        &self,
+        request: Request<SubmitRequest>,
+    ) -> Result<Response<SubmitResponse>, Status> {
+        debug!("received contract call request");
+        let request = request.get_ref();
+
+        let contract_call_data = match request.call_data.clone() {
+            Some(call) => call,
+            None => return Err(tonic::Status::invalid_argument("err")),
```
EricBolten: The string here could be more descriptive of what the error is.

steward/src/cork.rs
```diff
@@ -94,6 +105,140 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
...
+
+        let keystore = path::Path::new(&config.keystore);
+        let keystore = FsKeyStore::create_or_open(keystore).expect("Could not open keystore");
+
+        let name = &config
+            .keys
+            .delegate_key
+            .parse()
+            .expect("Could not parse name");
+        let key = keystore.load(name).expect("Could not load key");
```
EricBolten: You should be able to use `load_ethers_wallet` from `config` to replace a lot of this boilerplate.

steward/src/cork.rs
```diff
@@ -139,3 +284,11 @@ async fn send_cork(cork: Cork) -> Result<TxResponse, Error> {
     .await
     .map_err(|e| e.into())
 }
+
+async fn contract_call<T: Detokenize>(contract_call: ContractCall<EthSignerMiddleware, T>) {
+    let gas = contract_call.estimate_gas().await.unwrap();
+    let gas_price = CellarGas::get_gas_price().await.unwrap();
+    let contract_call = contract_call.gas(gas).gas_price(gas_price);
+    let contract_call = contract_call.send().await.unwrap();
```
EricBolten: We need to account for error conditions from executing this contract call, so we can't just `unwrap` it. We'll have to wait for it to resolve and return a `Result`.

steward/src/cork.rs
```diff
@@ -94,6 +105,140 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
...
+        let address: Address = cellar_address
+            .parse()
+            .expect("could not parse cellar address");
+
+        match contract_call_data {
+            AaveV2Stablecoin(call) => {
+                let contract = AaveV2StablecoinCellar::new(address, Arc::new(client));
+
+                match call.function.unwrap() {
+                    Function::EnterStrategy(_) => contract_call(contract.enter_strategy()).await,
```
EricBolten: Since these calls can fail, once `contract_call` is updated, the case where `await` returns an error result needs to be bubbled back up and returned as a `tonic` error to the caller.

steward/src/cork.rs
```diff
@@ -94,6 +105,140 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
     }
 }
 
+pub struct DirectCorkHandler;
```
EricBolten: There needs to be a CLI command similar to `steward/src/comands/start.rs` that uses the `DirectCorkHandler` instead of the `CorkHandler` so the server can be run in test mode.

EricBolten (commented):
Some suggestions.

steward/src/cork.rs
```diff
@@ -94,6 +108,121 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
...
+            None => {
+                return Err(tonic::Status::invalid_argument(
+                    "Error, can't find call data",
+                ))
+            }
+        };
+
+        let config = APP.config();
+
+        let name = &config.keys.delegate_key;
```
EricBolten: We've been using `delegate_key` to refer to the Cosmos key in the configuration. It's unlikely that someone is intending to use the same key material for their Ethereum testing. Maybe the Etheruem key to use needs to be an option on the command line.

steward/src/cork.rs
```diff
@@ -139,3 +268,14 @@ async fn send_cork(cork: Cork) -> Result<TxResponse, Error> {
...
 }
+
+async fn contract_call<T: Detokenize>(
+    contract_call: ContractCall<EthSignerMiddleware, T>,
+) -> Result<(), Error> {
+    let gas = contract_call.estimate_gas().await?;
+    let gas_price = CellarGas::get_gas_price().await?;
+    let contract_call = contract_call.gas(gas).gas_price(gas_price);
+    let contract_call = contract_call.send().await?;
+    let _tx_hash = *contract_call;
```
EricBolten: I don't think we want to re-assign `contract_call` here, something like `pending_tx` would be more appropriate. When we call `contract_call.send().await?`, we get back a pending transaction. It would be good to log that information so whoever is running the server knows that a transaction was submitted so they can e.g. look it up on Etherscan. The transaction hash is what you get when you dereference the pending transaction (the `*contract_call` bit in the current code).

After you get a pending transaction back, that's still a `Future`. You need to `await` it again for the transaction to fully complete.

steward/src/cork.rs
```diff
@@ -94,6 +108,121 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
...
+                    Function::TransferFees(_) => contract_call(contract.transfer_fees()).await,
+                    Function::SetLiquidityLimit(params) => {
+                        contract_call(contract.set_liquidity_limit(params.limit.into())).await
+                    }
+                    Function::ClaimAndUnstake(_) => {
+                        contract_call(contract.claim_and_unstake()).await
+                    }
+                }
+                .unwrap_or_else(|err| {
+                    panic!("{:?}", tonic::Status::invalid_argument(err.to_string()));
```
EricBolten: Using `panic!` here isn't helpful, we should be returning some kind of error back via the `Result` type of the `submit` function we're defining here. There are different `tonic` error types one could choose, here is an example of one: https://github.com/PeggyJV/steward/blob/1d761433f806ed9dbf4439432a04ca95590d4b55/steward/src/cork.rs#L86-L89

steward/src/cork.rs
```diff
@@ -94,6 +108,121 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
...
+                                    .map(|addr| match addr.parse::<H160>() {
+                                        Ok(addr) => Ok(addr),
+                                        Err(_) => Err(addr),
+                                    })
+                                    .collect::<Vec<_>>()
+                                    .iter()
+                                    .map(|r| r.unwrap())
+                                    .collect::<Vec<H160>>()
+                                    .try_into()
+                                    .expect("failed to convert 'route' addresses to array"),
```
EricBolten: Trying to do this all as one long statement directly as an argument to `rebalance` makes this a bit difficult to read, and also means you can't call `validate_route` on it like the other version is doing. It would be cleaner to define some intermediary variables to break up this long chain similar to how it's parsed in the existing code.

If we find that we are repeating a bunch of this code, perhaps it would be best to split it into a function in a single place that both can call.

EricBolten (commented):
Some comments.

steward/src/cork.rs
hannydevelop: Ooh, I thought we could just panic and display the `tonic::Status::invalid_argument` tonic error type.

steward/src/cork.rs
hannydevelop: done!

steward/src/cork.rs
hannydevelop: done!

steward/src/cork.rs
hannydevelop: done!

steward/src/commands/test_mode.rs
```diff
@@ -0,0 +1,69 @@
+use crate::{application::APP, config::StewardConfig, cork::DirectCorkHandler, prelude::*, server};
+use abscissa_core::{clap::Parser, config, Command, FrameworkError, Runnable};
+use std::result::Result;
+use steward_proto::steward::contract_call_server::ContractCallServer;
+
+/// Test mode, start Eth test mode
+#[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Test mode, start Ethereum test module.\n This command starts Steward using the Ethereum test mode."
```
EricBolten: This is really repetitive, we only need to say this is the test mode once.

steward/src/commands/test_mode.rs
```diff
@@ -0,0 +1,69 @@
...
+                });
+
+            info!("listening on {}", server_config.address);
+            if let Err(err) = tonic::transport::Server::builder()
+                .tls_config(server_config.tls_config)
+                .unwrap_or_else(|err| {
+                    panic!("{:?}", err);
+                })
+                .add_service(ContractCallServer::new(DirectCorkHandler {
+                    name: self.key_name.clone(),
```
EricBolten: `DirectCorkHandler` should also probably refer to this as `key_name` to avoid confusion.

steward/src/commands/test_mode.rs
```diff
@@ -0,0 +1,69 @@
...
+            std::process::exit(1)
+        });
+    }
+}
+
+impl config::Override<StewardConfig> for TestModeCmd {
+    // Process the given command line options, overriding settings from
+    // a configuration file using explicit flags taken from command-line
+    // arguments.
+    fn override_config(&self, config: StewardConfig) -> Result<StewardConfig, FrameworkError> {
```
EricBolten: Is this actually being used in this command?

steward/src/cork.rs
```diff
@@ -53,9 +69,9 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
             .cellar_ids
             .clone();
         if !ids.contains(&request.cellar_id) {
-            info!(
-                "rejecting request for unapproved cellar {}",
-                request.cellar_id
+            debug!(
```
EricBolten: Let's leave this as `info`.

steward/src/cork.rs
```diff
@@ -67,11 +83,10 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
         }
 
         // Build and send cork
-        let cellar_id = request.cellar_id.clone();
         let cork = match build_cork(request).await {
             Ok(c) => c,
             Err(err) => {
-                warn!("failed to build cork for cellar {}: {}", cellar_id, err);
+                warn!("failed to build cork: {}", err);
```
EricBolten: Let's leave this logging message as it was.

steward/src/cork.rs
```diff
@@ -84,11 +99,139 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
                 "failed to send cork to sommelier",
             ));
         }
-        info!("submitted cork for {}!", cellar_id);
+        debug!("sent cork!");
```
EricBolten: Same here, let's stick with the `info!` message.

steward/src/cork.rs
```diff
@@ -84,11 +99,139 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
...
+                        .await
+                    }
+                    Function::ClaimAndUnstake(_) => {
+                        contract_call(contract.claim_and_unstake()).await
+                    }
+                }
+            }
+        };
+        if function_call.is_err() {
+            return Err(Status::new(Code::Internal, "Function call error"));
```
EricBolten: Let's return the contents of the error so it's clear what error occurred.

steward/src/cork.rs
```diff
@@ -84,11 +99,139 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
...
+                            .route
+                            .iter()
+                            .map(|addr| match addr.parse::<H160>() {
+                                Ok(addr) => Ok(addr),
+                                Err(_) => Err(addr),
+                            })
+                            .collect();
+
+                        aave_v2_stablecoin::validate_route(results.clone()).unwrap_or_else(|err| {
+                            println!("{}", err);
```
EricBolten: If this fails we wouldn't want to proceed with the rest of the function and should return an error.

steward/src/cork.rs
```diff
@@ -84,11 +99,139 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
...
+                        .await
+                    }
+                    Function::Reinvest(params) => {
+                        contract_call(
+                            contract.reinvest(string_to_u256(params.min_assets_out).unwrap()),
+                        )
+                        .await
+                    }
+                    Function::Rebalance(params) => {
+                        let results: Vec<Result<H160, &String>> = params
```
EricBolten, line 171: If we're going to be calling this long list of validation functions in two places, perhaps we should this portion into a separate function in the Aave cellar code and then we could call that from here.

steward/src/cork.rs
```diff
@@ -141,3 +284,17 @@ async fn send_cork(cork: Cork) -> Result<TxResponse, Error> {
...
 }
+
+async fn contract_call<T: Detokenize>(
+    contract_call: ContractCall<EthSignerMiddleware, T>,
+) -> Result<(), Error> {
+    let gas = contract_call.estimate_gas().await?;
+    let gas_price = CellarGas::get_gas_price().await?;
+    let contract_call = contract_call.gas(gas).gas_price(gas_price);
+    let pending_tx = contract_call.send().await?;
+    let _tx_hash = *pending_tx;
```
EricBolten: We should print out an info message specifying the TX hash of the pending transaction while it is waiting to complete.

steward/src/cork.rs
```diff
@@ -141,3 +284,17 @@ async fn send_cork(cork: Cork) -> Result<TxResponse, Error> {
...
+    let transaction = pending_tx.await?;
+    info!("{:?}", transaction);
```
EricBolten: Might want to add some clarifying information to this log message to indicate this is a transaction receipt.

EricBolten (commented):
Some thoughts.

steward/src/commands/test_mode.rs
hannydevelop: I think it's safe to remove now since we're not overriding any configuration setting from command-line options

EricBolten:
We'll need to update this now that the contract has a new ABI.

cbrit:
Closing because of staleness

PR #122, opened 2022-04-20
wip for scheduled cork command

### LINKED ISSUE #97: Update steward with scheduled cork commands
Implement scheduled cork functions based on this table:

https://www.notion.so/Steward-Functions-222bb68963384616a8740c3645f73948

hannydevelop:
Closes #97


steward/src/commands.rs
```diff
@@ -31,6 +34,8 @@ pub const CONFIG_FILE: &str = "steward.toml";
 /// Steward Subcommands
 #[derive(Command, Debug, Parser, Runnable)]
 pub enum StewardCmd {
+    #[clap(subcommand)]
+    ScheduledCorks(ScheduledCorksCmd),
```
cbrit: Since it's an action we should probably call this `schedule-cork` or just `schedule` since the user doesn't need to know what a cork is.

steward/src/commands/scheduled_corks.rs
```diff
@@ -0,0 +1,13 @@
...
+
+use abscissa_core::{clap::Parser, Command, Runnable};
+
+/// Schedule corks commands
+#[derive(Command, Debug, Parser, Runnable)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Schedule corks for target contract.\n This command schedules cork by calling a function from the target Cellar. This is a validator only \n command and can only be run by validators. It Schedules cork based on the height specified by the \n validators. Therefore, it'll execute the function when the chain reaches that height."
+)]
+pub enum ScheduledCorksCmd {
+    Shutdown(Shutdown),
```
cbrit: I think we'll need a subcommand layer that specifies the cellar type because they may have different functions? Ideally the contract team decides on a standard for all cellars to have certain administrative functions but I don't think that's the case yet, so we'll need to be able to specify something like `steward schedule aave-v2-stablecoin shutdown <args>` for example.

steward/src/somm_send.rs
```diff
@@ -29,6 +29,23 @@ pub async fn send_cork(
...
+    delegate_key: &CosmosPrivateKey,
+    fee: Coin,
+    block_height: u64,
+) -> Result<TxResponse, CosmosGrpcError> {
+    let msg = MsgScheduleCorkRequest {
+        cork: Some(cork),
+        signer: delegate_address,
+        block_height,
+    };
+    let msg = Msg::new("/cork.v1.MsgSubmitCorkRequest", msg);
```
cbrit: this needs to be `"/cork.v1.MsgScheduleCorkRequest"`

steward/src/commands/scheduled_corks/shutdown.rs
```diff
@@ -0,0 +1,100 @@
...
+                downcast_to_u64(chain_id).expect("Chain ID overflowed when downcasting to u64");
+            let client = SignerMiddleware::new(provider, wallet.clone().with_chain_id(chain_id));
+
+            let address = self.contract;
+            let contract = AaveV2StablecoinCellar::new(address, Arc::new(client.clone()));
+            let call_data = match contract.shutdown().calldata() {
+                Some(call) => call,
+                None => return Err("Invalid Argument").expect("can't find call data"),
+            };
+            let encoded_call = call_data.to_vec();
```
cbrit: In this case we don't need the eth client/contract object because the generated abi code provides encoding without them:

```rust
use steward_abi::aave_v2_stablecoin::*;

let call = SetShutdownCall {
    shutdown: true,                     // these should be parameters in the command
    exit_position: true,
};
let encoded_call = AaveV2StablecoinCellarCalls::SetShutdown(call).encode();
```

More examples [here](https://github.com/PeggyJV/steward/blob/75e7469cb54bb86b20ff53a341f34d6847d4d8a4/steward/src/cellars/aave_v2_stablecoin.rs#L5)

steward/src/commands/scheduled_corks/shutdown.rs
```diff
@@ -0,0 +1,100 @@
...
+
+            let cork = Cork {
+                encoded_contract_call: encoded_call,
+                target_contract_address: address.to_string(),
```
cbrit: We'll want to do the same validation on this address that we do in the normal cork flow, i.e. check that it's [formatted correctly](https://github.com/PeggyJV/steward/blob/75e7469cb54bb86b20ff53a341f34d6847d4d8a4/steward/src/cellars.rs#L6-L12) and [that it's approved by governance](https://github.com/PeggyJV/steward/blob/75e7469cb54bb86b20ff53a341f34d6847d4d8a4/steward/src/cork.rs#L53-L72). I guess the formatting part is already taken care of since we are parsing it as an H160 from the start. For the second it would probably be good to pull that code into the `cellars` module as a method called `is_approved` or `check_cellar_approval` or something like that.

On that note I think we want to use `ethers::prelude::Address` rather than `H160`

steward/src/commands/scheduled_corks/shutdown.rs
```diff
@@ -0,0 +1,100 @@
...
+                amount: (cosmos_gas_price.0 as u64).into(),
+                denom: cosmos_gas_price.1,
+            };
+            somm_send::schedule_cork(
+                &contact,
+                cork,
+                config::DELEGATE_ADDRESS.to_string(),
+                &config::DELEGATE_KEY,
+                fee,
+                self.height,
```
cbrit: We should probably also do a query for the current block height to make sure this value isn't less than the current one.

steward/src/commands/scheduled_corks.rs
cbrit: The contract has been updated and this command is now called `SetShutdown` in the Aave cellar and has two parameters `shutdown: bool` and `exit_position: bool`:

https://github.com/PeggyJV/cellar-contracts/blob/b93b4393299aee35421f83ef4ca2689256e5b354/contracts/AaveV2StablecoinCellar.sol#L793

I have a PR open that will update the mock contract to reflect that

cbrit:
There may be a few other commands we need than set_shutdown. We'll need to nail down what they are.

hannydevelop:
> There may be a few other commands we need than set_shutdown. We'll need to nail down what they are.

Yes. I'm thinking the "pause" command too. @EricBolten wanted us to start with shutdown first

steward/src/somm_send.rs
hannydevelop: Oops. I missed that

steward/src/commands/scheduled_corks.rs
hannydevelop: Noted. You'll have to get that merged if that be the case

steward/src/commands.rs
hannydevelop: I think the name fits well. It's `scheduled-corks`

steward/src/commands/scheduled_corks.rs
hannydevelop: Yup. That's something we could do in future. However, I'll like to keep it simple now

steward/src/commands/scheduled_corks/shutdown.rs
hannydevelop: Ok

steward/src/commands/scheduled_corks/shutdown.rs
hannydevelop: Noted

steward/src/commands.rs
EricBolten: In retrospect I kind of agree here, I think `schedule` probably works better. I don't think we directly expose the cork abstraction elsewhere.

steward/src/commands/scheduled_corks.rs
EricBolten: I think this distinction is actually pretty important, let's add this make it clear and to avoid having to change the command structure in the future.

steward/src/commands/scheduled_corks/shutdown.rs
```diff
@@ -0,0 +1,100 @@
...
+use somm_proto::cork::Cork;
+use std::{convert::TryFrom, path, sync::Arc, time::Duration};
+use steward_abi::aave_v2_stablecoin::AaveV2StablecoinCellar;
+const MESSAGE_TIMEOUT: Duration = Duration::from_secs(10);
+const CHAIN_PREFIX: &str = "somm";
+
+/// Shutdown subcommand
+#[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Shutdown target Cellar.\n This command schedules the shutdown of a Cellar. This is a validator only \n command and can only be run by validators. It Schedules shutdown based on the height specified by \n the validators. Therefore, it'll execute the function when the chain reaches that height."
```
EricBolten: We probably don't need to repeat all the info from the `schedule` parent command in the long description here.

steward/src/commands/scheduled_corks/shutdown.rs
EricBolten: Yes, we don't need to create an eth client to get the encoded call.

steward/src/commands/scheduled_corks/shutdown.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// Block height to schedule cork.
+    #[clap(short = 'b', long)]
+    height: u64,
+}
+
+impl Runnable for Shutdown {
+    fn run(&self) {
+        let config = APP.config();
+
+        let keystore = path::Path::new(&config.keystore);
```
EricBolten: Let's remove all the ethereum setup boilerplate after switching call encoding methods below.

steward/src/commands/scheduled_corks/shutdown.rs
EricBolten: We'll actually want to take this in as a string. You can validate that the string is a valid Ethereum address, but `Cork` is expecting a string type. Calling `to_string` on an `H160` or `Address` provides weird truncated output without manually converting it into bytes and outputting the hex, we ran into this in the Gravity Bridge as well. Can be solved by just accepting the arg as a string input.

steward/src/commands/scheduled_corks/shutdown.rs
```diff
@@ -0,0 +1,100 @@
...
+            somm_send::schedule_cork(
+                &contact,
+                cork,
+                config::DELEGATE_ADDRESS.to_string(),
+                &config::DELEGATE_KEY,
+                fee,
+                self.height,
+            )
+            .await
+            .expect("err");
```
EricBolten: Could use a more detailed message here.

EricBolten (commented):
Some suggestions and comments.

steward/src/commands/scheduled_corks.rs
hannydevelop: done!

steward/src/somm_send.rs
hannydevelop: done!

steward/src/commands/scheduled_corks/shutdown.rs
hannydevelop: done!

steward/src/commands/scheduled_corks/shutdown.rs
hannydevelop: all good!

steward/src/commands/schedule_corks/aave_cellar/shutdown.rs
```diff
@@ -0,0 +1,78 @@
...
+#[clap(
+    long_about = "DESCRIPTION \n\n Shutdown target Cellar.\n This command schedules the shutdown of a Cellar. This is a validator only \n command and can only be run by validators. It Schedules shutdown based on the height specified by \n the validators. Therefore, it'll execute the function when the chain reaches that height."
+)]
+pub struct ShutdownCmd {
+    /// Target contract for scheduled cork.
+    #[clap(short, long)]
+    contract: String,
+
+    /// Block height to schedule cork.
+    #[clap(short = 'b', long)]
```
EricBolten, line 28: Probably want `h` as the short code here for `height`.

steward/src/commands/schedule_corks/aave_cellar/shutdown.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+impl Runnable for ShutdownCmd {
+    fn run(&self) {
+        let config = APP.config();
+
+        abscissa_tokio::run_with_actix(&APP, async {
+            let call = SetShutdownCall {
+                shutdown: true,
+                exit_position: true,
```
EricBolten: We should allow the caller to specify these parameters. The shutdown function can be called with true or false values depending on what the intention is. The current Aave contract doesn't have being shutdown as a permanent condition, it's more like a pause.

steward/src/commands/schedule_corks/aave_cellar/shutdown.rs
```diff
@@ -0,0 +1,78 @@
...
+            };
+            let encoded_call = AaveV2StablecoinCellarCalls::SetShutdown(call).encode();
+
+            cellars::validate_cellar_id(self.contract.as_str()).unwrap_or_else(|err| {
+                status_err!("Can't validate cellar ID: {}", err);
```
EricBolten: We should probably change this message to something like "can't validate contract address format" since the command isn't really calling them cellar IDs.

steward/src/commands/schedule_corks/aave_cellar/shutdown.rs
```diff
@@ -0,0 +1,78 @@
...
+            somm_send::schedule_cork(
+                &contact,
+                cork,
+                config::DELEGATE_ADDRESS.to_string(),
+                &config::DELEGATE_KEY,
+                fee,
+                self.height,
+            )
+            .await
+            .expect("Error, can't schedule shutdown");
```
EricBolten: This will return a `CosmosGrpcError` indicating what the problem is, it would be good to capture that info rather than just printing that there was an error.

EricBolten (commented):
Some comments.

steward/src/commands/schedule_corks/aave_cellar/shutdown.rs
hannydevelop: The reason the `block_height`'s shortcode is `b` is because `h` is taken for `help` in all of Abscissa's commands. If I go-ahead to add `h` for `block_height`, I'll get this error:

```
thread 'main' panicked at 'App shutdown: Short option names must be unique for each argument, but '-h' is in use by both 'height' and 'help'', /home/ugochi/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-3.0.12/src/build/app/debug_asserts.rs:100:17
```

steward/src/commands/schedule_corks/aave_cellar/shutdown.rs
hannydevelop: ooh, I thought this will always be true. Added!

steward/src/commands/schedule_corks/aave_cellar/shutdown.rs
hannydevelop: done!

steward/src/commands/schedule_corks/aave_cellar/shutdown.rs
EricBolten: 👍

steward/src/commands/schedule_corks/aave_cellar/shutdown.rs
```diff
@@ -0,0 +1,88 @@
...
+                &contact,
+                cork,
+                config::DELEGATE_ADDRESS.to_string(),
+                &config::DELEGATE_KEY,
+                fee,
+                self.height,
+            )
+            .await
+            .unwrap_or_else(|err| {
+                panic!("{:?}", err);
```
EricBolten: Rather than using `panic!` here we should exit the process cleanly in the same way we are handling other error conditions: printing a message with `status_err!` and then using `exit(1)`.

EricBolten (commented):
A final comment.

EricBolten (approved):
LGTM

PR #125, opened 2022-04-26
Rust reference client

philipjames44:
Grpc Reference Client that allows a user to specify a cosmos endpoint, steward endpoints, and TLS configurations for connections.

Client blocks until active voting period, then sends requests in parallel to all steward instances specified before returning a tuple of results and failures.


steward/src/grpc_client.rs
```diff
@@ -0,0 +1,158 @@
...
+                .into());
+        }
+
+        Ok(())
+    }
+
+    pub fn check_for_steward_addresses(&self) -> Result<(), Error> {
+        if self.steward_endpoints.is_empty() {
+            return Err(ErrorKind::GrpcError
+                .context("no grpc address available")
```
hannydevelop: I think this should be `No endpoint found` or `steward address unavailable`. Since we already have the `no grpc address available` message, I think duplicating it is a bad idea.

steward/src/grpc_client.rs
```diff
@@ -0,0 +1,158 @@
...
+            };
+
+            if commit_period_response.current_height >= commit_period_response.vote_period_start
+                && commit_period_response.current_height < commit_period_response.vote_period_end
+            {
+                break;
+            }
+
+            // Wait a few seconds before retrying
+            thread::sleep(Duration::from_secs(10));
```
hannydevelop: Don't you think it'll be good if we can use `tokio::time` here so that we can make a timeout for our loop? Something like this: 

https://github.com/PeggyJV/gravity-bridge/blob/16bf5569ec490fe333b99cc36231021a3b96a134/orchestrator/gorc/src/commands/deploy/erc20.rs#L103

philipjames44: That's a good idea, I'll add that in

steward/src/grpc_client.rs
philipjames44: Agreed, that'd make it clearer, will update

cbrit:
Closing because of staleness

PR #127, opened 2022-04-26
Steward setup checklist

cbrit:


docs/02-StewardForValidators.md
```diff
@@ -16,10 +16,21 @@ steward -c <config_toml_path> start
...
 
 ## Running Steward as a server
 
+### Setup Checklist
+
+- [ ] Domain name for Strategy Providers to use for requests to Steward
+- [ ] Firewall configured to allow requests to hit the [Steward port](./01-Configuration.md#port) over the internet
+- [ ] Connectivity between Steward and your Sommelier node's gRPC endpoint
+- [ ] On-disk key store containing a [delegate key](./02-StewardForValidators.md#cosmos-delegate-key) in this key store for Steward and Orchestrator to share. This can be generated with Steward, or you may have already done this step with `gorc` in the past. If that's the case you can just use the delegate key you’ve already created.
+- [ ] A TLS 1.3 [server CA and certificate signed with said CA](./02-StewardForValidators.md#tls-certificates)
```
EricBolten: This section doesn't contain instructions yet for creating them, so perhaps we should wait to add this. The other open PR regarding that has some feedback in it.

docs/02-StewardForValidators.md
```diff
@@ -16,10 +16,21 @@ steward -c <config_toml_path> start
...
+- [ ] [TOML config file](./01-Configuration.md#complete-example-configtoml) containing the above values
+- [ ] Add your Steward info (endpoint and CA) to the [Steward registry repository](https://github.com/peggyjv/steward-registry#steward-registry)
```
EricBolten, line 33: So being in this liminal space before the next upgrade to put this data on-chain is annoying, but if we're going to do this we'll need admins on this repo capable of merging to follow up with validator representatives to confirm the PR is really from them, and that the domain they're using is their validators domain. Social proof for this bit is going to be annoying but necessary. When we introduce pubsub validators will be able to set their domain and cert using their delegate key.

EricBolten (commented):
Some comments.

EricBolten (approved):
LGTM

PR #128, opened 2022-04-27
Collin/update aave api

### LINKED ISSUE #105: Strategy Provider API documentation
Write or (preferably) generate documentation of the proto messages for function calls available to SPs.

### LINKED ISSUE #126: Update AaveV2Stablecoin cellar API and mock contract update

cbrit:
Closes #126
Closes #105


0xble:
Looks like `docs/api/aave_v2_stablecoin.md` still need to be updated?

cbrit:
> Looks like `docs/api/aave_v2_stablecoin.md` still need to be updated?

which part? it was the last thing I generated

docs/api/aave_v2_stablecoin.md
```diff
@@ -0,0 +1,199 @@
...
+
+
+
+<a name="steward-v1-AaveV2Stablecoin"></a>
+
+### AaveV2Stablecoin
+Represents a function call to the Aave V2 Stablecoin cellar
+
+
+| Field | Type | Label | Description |
```
hannydevelop, line 35: If the `label` and `description` fields are going to be empty, it'll be best to leave them out.

docs/api/aave_v2_stablecoin.md
```diff
@@ -0,0 +1,199 @@
...
+
+
+
+
+<a name="steward-v1-AaveV2Stablecoin-AccrueFees"></a>
+
+### AaveV2Stablecoin.AccrueFees
+Take platform fees and performance fees off of cellar&#39;s active assets.
+
+Represents function `accruePlatformFees()``
```
hannydevelop: Remove extra tick from `accruePlatformFees()`

hannydevelop: Overall, I think the `label` field should be expunged from the docs since it's empty

cbrit: Label is used in a couple of place, not much though. These are auto-generated so I'm not sure it's a good idea to edit them manually since we'd have to do it every time we make changes to the proto or proto comments.

docs/api/aave_v2_stablecoin.md
cbrit: I should fill in the descriptions there

integration_tests/ethereum/contracts/MockAaveV2StablecoinCellar.sol
```diff
@@ -5,109 +5,129 @@ import "./interfaces.sol";
...
+    event mockSetTrust(address position, bool trust);
+    event mockSetShutdown(bool shutdown, bool existPosition);
+
+    constructor() {}
 
     /**
      * @notice Take platform fees and performance fees off of cellar's active assets.
      */
-    function accrueFees() external onlyOwner {
+    function accrueFees() external {
```
philipjames44, line 30: Why remove the onlyOwner modifier?

integration_tests/ethereum/contracts/MockAaveV2StablecoinCellar.sol
```diff
@@ -5,109 +5,129 @@ import "./interfaces.sol";
...
+    event mockSetFeesDistributor(bytes32 newFeesDistributor);
+    event mockEnterPosition();
+    event mockRebalance(address[9] route, uint256[3][4] swapParams, uint256 minAssetsOut);
+    event mockReinvest(uint256 minAssetsOut);
+    event mockClaimAndUnstake();
+    event mockSweep(address token, address to);
+    event mockSetLiquidityLimit(uint256 limit);
+    event mockSetDepositLimit(uint256 limit);
+    event mockSetTrust(address position, bool trust);
+    event mockSetShutdown(bool shutdown, bool existPosition);
```
EricBolten: `existPosition` -> `exitPosition`

integration_tests/ethereum/contracts/MockAaveV2StablecoinCellar.sol
EricBolten: This is reflecting what the current contract is doing. I think `accrueFees` is non-dangerous so I assume the theory is that if someone else wants to pay for it that's fine.

docs/api/aave_v2_stablecoin.md
EricBolten: We shouldn't muck with the autogenerated files unless there's a way to adjust the generator's configuration.

steward/src/cellars/aave_v2_stablecoin.rs
```diff
@@ -1,77 +1,182 @@
...
+            }
+
+            if params.swap_params.len() != 4 {
+                return Err(sp_call_error(format!(
+                    "{}: Rebalance 'swap_params': array must contain 4 elements",
+                    LOG_PREFIX
+                )));
+            }
+
+            let results: Vec<Result<H160, &String>> = params
```
EricBolten, line 44: If these `H160` elements are meant to be addresses, we should use the `Address` type.

integration_tests/ethereum/contracts/MockAaveV2StablecoinCellar.sol
```diff
@@ -5,109 +5,129 @@ import "./interfaces.sol";
...
+                         Polygon factory metapools `exchange_underlying`
+     * @param minAssetsOut minimum amount of assets received from swap
      */
-    function rebalance(address[] memory path, uint256 amountOutMinimum) external onlyOwner  {
-        emit mockRebalance(path, amountOutMinimum);
+    function rebalance(
+        address[9] memory route,
+        uint256[3][4] memory swapParams,
+        uint256 minAssetsOut
+    ) external onlyOwner  {
```
EricBolten, line 71: To address concerns I've seen you raise around handling the input correctly, should we add some code to the mock contract to confirm the input format is correct before emitting the event?

steward/src/cellars/aave_v2_stablecoin.rs
```diff
@@ -1,77 +1,182 @@
...
-                    .parse::<H160>()
-                    .expect("failed to parse token address"),
-                is_approved: params.is_approved,
-            };
-            AaveV2StablecoinCellarCalls::SetInputToken(call).encode()
+    }
+}
+
+fn validate_route(results: Vec<Result<H160, &String>>) -> Result<(), Error> {
+    let mut bad_addresses = String::new();
```
EricBolten: Did you want this to be a `Vec<String>`?

steward/src/cellars/aave_v2_stablecoin.rs
```diff
@@ -1,77 +1,182 @@
...
+    for r in results {
+        if let Err(addr) = r {
+            bad_addresses.push_str(&format!(", {}", addr))
```
EricBolten: This method has the weird side effect of starting the line with a comma.

steward/src/cellars/aave_v2_stablecoin.rs
```diff
@@ -1,77 +1,182 @@
...
+                LOG_PREFIX,
+                sp
+            )));
         }
     }
+
+    Ok(())
+}
+
+fn sp_call_error(message: String) -> Error {
```
EricBolten: We've defined this function here in a cellar-specific file, but we also manually call `ErrorKind::SPCallError.context().into()` elsewhere, do we want this helped function moved and more broadly used?

steward/src/cellars/aave_v2_stablecoin.rs
```diff
@@ -1,77 +1,182 @@
...
+    if !bad_addresses.is_empty() {
+        let mut err_string = "Rebalance 'route': array contains invalid address(s)".to_string();
+        err_string.push_str(&bad_addresses);
+        return Err(sp_call_error(format!("{}: {}", LOG_PREFIX, err_string)));
+    }
+
+    Ok(())
+}
+
+fn validate_swap_params(swap_params: Vec<SwapParams>) -> Result<(), Error> {
```
EricBolten: Can we add some test cases for these two params validation functions?

This is also amplifying the fact that in the future adding more cellars this way requires a decent amount of code. Probably going to take some deep thinking to figure out a more elegant solution.

steward_proto/proto/aave_v2_stablecoin.proto
```diff
@@ -1,68 +1,118 @@
...
-        Reinvest reinvest = 3;
-        ClaimAndUnstakeWithAmount claim_and_unstake_amount = 4;
-        ClaimAndUnstake claim_and_unstake = 5;
-        Rebalance rebalance = 6;
-        AccruePlatformFees accrue_platform_fees = 7;
-        TransferFees transfer_fees = 8;
-        SetInputToken set_input_token = 9;
-        RemoveLiquidityRestriction remove_liquidity_restriction = 10;
-        Sweep sweep = 11;
+        // Represents function `accruePlatformFees()`: [link](https://github.com/PeggyJV/cellar-contracts/blob/b93b4393299aee35421f83ef4ca2689256e5b354/contracts/AaveV2StablecoinCellar.sol#L702)
```
EricBolten: Did you manually add these? The code it's pointing to is point-in-time based on a commit...if say the arguments of the function change, you'd have to know that and update this comment by hand. Something to consider whether or not it's a valuable addition, perhaps just linking to the overall contract at the top would suffice.

steward_proto/proto/aave_v2_stablecoin.proto
```diff
@@ -1,68 +1,118 @@
...
-
-// function accruePlatformFees() external;
-message AccruePlatformFees {}
-
-// function transferFees() external;
-message TransferFees {}
+        /*
+         * Represents parameters for a single swap. Each swap needs the indeces in Rebalance.route of the in/out token addresses and the swap type. See the Curve contract linked above for more detail.
+         */
+        message SwapParams {
```
EricBolten, line 76: It kind of feels ugly our code has to understand the semantics of the particular arguments in order to be able to to parse it and convert it into the expected format by the contract, but there isn't much alternative at the moment absent requiring the SP to format the encoded call and doing inspection of that call rather than creating it. Something to think about.

EricBolten (commented):
Some comments.

docs/api/aave_v2_stablecoin.md
cbrit: > We shouldn't muck with the autogenerated files unless there's a way to adjust the generator's configuration.

added comments so it would generate descriptions

integration_tests/ethereum/contracts/MockAaveV2StablecoinCellar.sol
cbrit: Does that give us any more info that validating in steward does?

EricBolten: The integration test is meant to be testing whether or not the calls generated by steward are hitting the contract correctly. We're only using a mock contract because using the real one is infeasible.

steward/src/cellars/aave_v2_stablecoin.rs
cbrit: It get's appending to `err_string` so the comma ends up making some sense. would prefer the first be a `:` but it still looks ok:

```bash
ERROR:
  Code: InvalidArgument
  Message: SP call error: AaveV2StablcoinCellar: Rebalance 'route' argument contained invalid address(s), 0x5FC8d32690cc91D4c39d9d3abcBD16989F8757071, 0x0000000000000000000000000000000000000
```

steward/src/cellars/aave_v2_stablecoin.rs
cbrit: Yes but I figured it would need to be a method per error variant otherwise it becomes almost as verbose as what it's replacing. I do think it would make more sense to house it in `error.rs` or `utils.rs`.

steward/src/cellars/aave_v2_stablecoin.rs
cbrit: Will do. And yes great point...

steward_proto/proto/aave_v2_stablecoin.proto
cbrit: Yeah I was solving the other side of the problem which is linking to a specific line of the head of `main` which could change. Will link at the top.

steward_proto/proto/aave_v2_stablecoin.proto
cbrit: We could expect SPs to do input validation themselves, it kind of makes sense in the case where they are the ones who created the cellar. I wanted to make the UX and validation simpler to avoid wasted gas on mistakes but in reality we probably don't want to be in the business of validating all args... If we put that responsibility on the SP then it would reduce the amount of code we need per cellar as you mentioned elsewhere.

steward/src/cellars/aave_v2_stablecoin.rs
cbrit: The ABI binding actually expects an `H160` for the `RebalanceCall` struct to get encoded.

EricBolten: The `ethers` `Address` type is an `H160`, it just provides a bunch of support functions for being an address rather than just a sized hash.

steward/src/cellars/aave_v2_stablecoin.rs
cbrit: No, lines 48-55 return bad addresses as the `&String`s for logging. Keeping it in this form seemed to be the cleanest way to validate with one log message and continue:

https://github.com/PeggyJV/steward/blob/8a2209a8cf456a62d2ce7d21de27541336c4a022/steward/src/cellars/aave_v2_stablecoin.rs#L48-L65

cbrit: No, renaming to `bad_addresses_string`

integration_tests/cork_test.go
```diff
@@ -112,8 +112,8 @@ func (s *IntegrationTestSuite) TestCork() {
 					CellarId: cellarId,
 					CallData: &SubmitRequest_AaveV2Stablecoin{
 						&AaveV2Stablecoin{
-							Function: &AaveV2Stablecoin_ClaimAndUnstake{
-								&ClaimAndUnstake{},
+							Function: &AaveV2Stablecoin_ClaimAndUnstake_{
+								&AaveV2Stablecoin_ClaimAndUnstake{},
```
EricBolten, line 116: Now that we have this integration test functioning well, perhaps we should add some of the more complicated calls that have some arguments to ensure things are working as we expect.

steward_proto/proto/aave_v2_stablecoin.proto
```diff
@@ -1,68 +1,118 @@
...
     oneof function {
-        EnterStrategy enter_strategy = 1;
-        ReinvestWithAmount reinvest_amount = 2;
-        Reinvest reinvest = 3;
-        ClaimAndUnstakeWithAmount claim_and_unstake_amount = 4;
-        ClaimAndUnstake claim_and_unstake = 5;
-        Rebalance rebalance = 6;
-        AccruePlatformFees accrue_platform_fees = 7;
-        TransferFees transfer_fees = 8;
-        SetInputToken set_input_token = 9;
```
EricBolten, line 16: For `SetInputToken`, `RemoveLiquidityRestriction`, etc, there are new functions (`SetTrust`, `SetLiquidityLimit`, `SetDepositLimit`, etc). Have you checked with the data team as to whether or not these should be callable?

josephterrigno: I believe we want `SetTrust`, `SetLiquidityLimit`, `SetDepositLimit`... to only be callable by validators, through scheduled corks.

cbrit: > I believe we want `SetTrust`, `SetLiquidityLimit`, `SetDepositLimit`... to only be callable by validators, through scheduled corks.

This was my impression as well as these are administrative tasks.

steward/src/cellars/aave_v2_stablecoin.rs
```diff
@@ -1,77 +1,143 @@
...
-        ReinvestAmount(params) => {
-            let call = ReinvestWithAmountCall {
-                amount: params.amount.into(),
+        EnterPosition(_) => {
+            let call = EnterPositionCall {};
+            let call = AaveV2StablecoinCellarCalls::EnterPosition(call);
+            Ok(call.encode())
+        }
+        Rebalance(params) => {
+            // We expect the client to pad the route to length 9
```
EricBolten, line 29: If we expect the client to both pad out the `route` and `swap_params`, is the contract expecting zero values then to represent "don't do a swap"?

steward/src/cellars/aave_v2_stablecoin.rs
```diff
@@ -1,77 +1,143 @@
...
+            let call = SetLiquidityLimitCall {
+                limit: params.limit.into(),
             };
-            AaveV2StablecoinCellarCalls::Rebalance(call).encode()
+            Ok(AaveV2StablecoinCellarCalls::SetLiquidityLimit(call).encode())
         }
-        AccruePlatformFees(_) => {
-            let call = AccruePlatformFeesCall {};
-            AaveV2StablecoinCellarCalls::AccruePlatformFees(call).encode()
+        Sweep(params) => {
```
EricBolten: I think we should leave sweep out of the SP functions for now. We can add it back later if we think SPs should be able to call it.

steward/src/cellars/aave_v2_stablecoin.rs
cbrit: It stops processing when it reaches a zero value yes. There isn't a situation where you'd want zero swaps afaik.

EricBolten (approved):
LGTM

PR #129, opened 2022-05-03
Use Rebalance function in test

cbrit:


integration_tests/cork_test.go
```diff
@@ -152,8 +169,12 @@ func (s *IntegrationTestSuite) TestCork() {
 		aave_abi, err := AaveV2MetaData.GetAbi()
 		s.Require().NoError(err)
 
-		methodName := "claimAndUnstake"
-		method, err := aave_abi.Pack(methodName)
+		methodName := "rebalance"
+		route := [9]common.Address{common.HexToAddress(dai), common.HexToAddress(pool), common.HexToAddress(usdc), common.HexToAddress(zeroAddr), common.HexToAddress(zeroAddr), common.HexToAddress(zeroAddr), common.HexToAddress(zeroAddr), common.HexToAddress(zeroAddr), common.HexToAddress(zeroAddr)}
```
EricBolten: You don't really need to call `comon.HexToAddress(zeroAddr)` over and over again, you could just assign it as a variable.

integration_tests/cork_test.go
```diff
@@ -272,7 +293,15 @@ func (s *IntegrationTestSuite) TestCork() {
...
 			if len(logs) == 1 {
 				s.T().Log("saw mock function event!")
-				return true
+
+				log := logs[0]
+				if len(log.Data) > 0 {
+					var event AaveV2MockRebalance
+					err := aave_abi.UnpackIntoInterface(&event, "mockRebalance", log.Data)
+					s.Require().NoError(err, "failed to unpack mockRebalance event from log data")
+					s.Require().Equal(event.MinAssetsOut, big.NewInt(1000))
```
EricBolten, line 303: If you're going to inspect `MinAssetsOut`, is it worth also checking the route and swap values?

EricBolten (commented):
Some comments.

integration_tests/cork_test.go
cbrit: Seemed unnecessarily complicated to me?

integration_tests/cork_test.go
cbrit: ah meant to fix this

EricBolten (approved):
LGTM

PR #137, opened 2022-05-11
Build release without using `cross`

EricBolten:
It seems like using the `cross` tool is breaking the vendoring in the way the dependency tree is structured for steward vs. gravity bridge. This version of the release action targets `x86_64-unknown-linux-musl` directly using `cargo build` and pins to the same openssl crate versions as gravity bridge.


hannydevelop (approved):

hannydevelop:
This looks good. Bye bye `cross`

PR #144, opened 2022-05-13
Changed localhost -> fqdn

Tomas-Eminger:
The certificate itself is not for localhost, but for FQDN of the server


cbrit (approved):
Thanks Tomas. I wrote this assuming users would be running steward on the same box but that's a bad assumption.

PR #145, opened 2022-05-17
Make relaying opt-in

cbrit:
To help prevent validators from accidentally running the relayer and potentially spending ETH for gas if their wallets are funded.

Changes the flag from `--orchestrator-only` to `--relay`

### DIFF
--- a/steward/src/commands/orchestrator/start.rs
+++ b/steward/src/commands/orchestrator/start.rs
@@ -16,10 +16,10 @@ use gravity_bridge::orchestrator::main_loop::{
 use gravity_bridge::relayer::main_loop::LOOP_SPEED as RELAYER_LOOP_SPEED;
 use std::{cmp::min, sync::Arc};
 
-/// Start Orchestrator
+/// Start the Orchestrator
 #[derive(Command, Debug, Parser)]
 #[clap(
-    long_about = "DESCRIPTION \n\n Start the Orchestrator in Sommelier Chain via the Gravity Bridge.\n This command loads a Cosmos and Ethereum key with their keyname from the keystore.\n It also takes an Orchestrator_only field which when set to true, starts the Orchestrator only\n without the relayer and when set to false, starts the Orchestrator with the relayer."
+    long_about = "DESCRIPTION\n\nStarts the Gravity Bridge Orchestrator and optionally the Relayer with the --relay flag.\nYou must provide the keys delegated by your validator node for Cosmos and Ethereum signing."
 )]
 pub struct StartCommand {
     /// Cosmos keyname from keystore.
@@ -30,7 +30,7 @@ pub struct StartCommand {
     #[clap(short = 'e', long)]
     ethereum_key: String,
 
-    /// Boolean, when set to true starts the Orchestrator only and false starts Orchestrator and Relayer.
+    /// Enable Ethereum to Cosmos relaying (consumes gas in ETH)
     #[clap(short, long)]
     relay: bool,
 }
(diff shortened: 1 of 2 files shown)


philipjames44:
Will also require some doc updates https://github.com/PeggyJV/steward/blob/main/docs/03-TheOrchestrator.md?plain=1#L9-L13

steward/src/commands/orchestrator/start.rs
```diff
@@ -32,7 +32,7 @@ pub struct StartCommand {
```
philipjames44, line 32: Believe we want to update the 'Orchestrator_only' language on line 22 above as well

hannydevelop (commented):
In my opinion, the orchestrator only flag was better. We have help texts that display and give further information on the flags and overall command. Also, `!self.relay` becomes double negative when the user chooses `false`.

However, if we're changing this, you'll need to update the description at the beginning of the commands so they don't get further confused

cbrit:
> In my opinion, the orchestrator only flag was better. We have help texts that display and give further information on the flags and overall command. Also, `!self.relay` becomes double negative when the user chooses `false`.
> 
> However, if we're changing this, you'll need to update the description at the beginning of the commands so they don't get further confused

We've seen a few validators starting the relayer by accident because it is default. Since we only want a couple of validators running the relayer, this switch will make it harder to do accidentally.

Will update the help/description!

steward/src/commands/orchestrator/start.rs
```diff
@@ -16,10 +16,10 @@ use gravity_bridge::orchestrator::main_loop::{
 use gravity_bridge::relayer::main_loop::LOOP_SPEED as RELAYER_LOOP_SPEED;
 use std::{cmp::min, sync::Arc};
 
-/// Start Orchestrator
+/// Start the Orchestrator
 #[derive(Command, Debug, Parser)]
 #[clap(
-    long_about = "DESCRIPTION \n\n Start the Orchestrator in Sommelier Chain via the Gravity Bridge.\n This command loads a Cosmos and Ethereum key with their keyname from the keystore.\n It also takes an Orchestrator_only field which when set to true, starts the Orchestrator only\n without the relayer and when set to false, starts the Orchestrator with the relayer."
+    long_about = "DESCRIPTION\n\nStarts the Gravity Bridge Orchestrator and optionally the Relayer with the --relay flag.\nYou must provide the keys delegated by your validator node for Cosmos and Ethereum signing."
```
EricBolten, line 22: With the adjusted language from the original, we should be specific that we mean key names and not the keys themselves.

steward/src/commands/orchestrator/start.rs
```diff
@@ -118,7 +118,7 @@ impl Runnable for StartCommand {
                 config.ethereum.gas_price_multiplier,
                 config.ethereum.blocks_to_search,
                 config.cosmos.gas_adjustment,
-                self.orchestrator_only,
+                !self.relay,
```
EricBolten, line 121: It's a little awkward to essentially use a double negative here, and we should implement this change across both gorc and steward. Perhaps we should first make this update to gorc in both the CLI and adjusting the orchestrator to expect a true value here to relay, update steward with that new version, and then change the behavior in steward.

EricBolten (commented):
Some suggestions.

steward/src/commands/orchestrator/start.rs
cbrit: Ugochi thought the same thing. I'll update gorc.

EricBolten:
I don't believe we've followed up on the last comments here, pinging to check in.

cbrit:
Closing because of staleness

PR #147, opened 2022-05-23
Commands for all scheduled cork functions

### LINKED ISSUE #97: Update steward with scheduled cork commands

hannydevelop:
closes #97


steward/src/commands/schedule_corks/aave_cellar/fees_distributor.rs
```diff
@@ -0,0 +1,92 @@
...
+    somm_send,
+};
+use abscissa_core::{clap::Parser, Command, Runnable};
+use deep_space::{Coin, Contact};
+use ethers::abi::AbiEncode;
+use somm_proto::cork::Cork;
+use std::time::Duration;
+use steward_abi::aave_v2_stablecoin::*;
+
+const MESSAGE_TIMEOUT: Duration = Duration::from_secs(10);
```
EricBolten: These constants should probably live elsewhere as they will apply to all scheduled corks.

steward/src/commands/schedule_corks/aave_cellar/fees_distributor.rs
```diff
@@ -0,0 +1,92 @@
...
+const CHAIN_PREFIX: &str = "somm";
+
+/// Fees Distributor subcommand
+#[derive(Command, Debug, Parser)]
+pub struct FeesDistributorCmd {
+    #[clap(short = 'n', long)]
+    new_fees_distributor: String,
```
EricBolten: This could use a descriptive comment for this flag.

steward/src/commands/schedule_corks/aave_cellar/fees_distributor.rs
```diff
@@ -0,0 +1,92 @@
...
+    height: u64,
+}
+
+impl Runnable for FeesDistributorCmd {
+    fn run(&self) {
+        let config = APP.config();
+
+        abscissa_tokio::run_with_actix(&APP, async {
+            let mut bytes = [0u8; 32];
+            if let Err(err) = hex::decode_to_slice(self.new_fees_distributor.clone(), &mut bytes) {
```
EricBolten: As discussed out of band, this will need to be formatted in the same way the orchestrator handles formatting Cosmos addresses.

steward/src/commands/schedule_corks/aave_cellar/fees_distributor.rs
```diff
@@ -0,0 +1,92 @@
...
+                error!("failed to decode to slice: {}", err);
+            }
+            let call = SetFeesDistributorCall {
+                new_fees_distributor: bytes,
+            };
+
+            let encoded_call = AaveV2StablecoinCellarCalls::SetFeesDistributor(call).encode();
+
+            // Validate cellar id
+            cellars::validate_cellar_id(self.contract.as_str()).unwrap_or_else(|err| {
```
EricBolten: Everything from the `valiate_cellar_id` call to the final `schedule_cork` call is repeated verbatim for every scheduled cork. We should extract this into a common function.

steward/src/commands/schedule_corks/aave_cellar/trust.rs
```diff
@@ -0,0 +1,96 @@
...
+/// Trust subcommand
+#[derive(Command, Debug, Parser)]
+
+pub struct TrustCmd {
+    /// Asset position
+    #[clap(short = 'p', long)]
+    position: H160,
+
+    #[clap(short = 't', long)]
+    trust: bool,
```
EricBolten: This flag could use a descriptive comment as well.

EricBolten (commented):
Some suggestions.

steward/src/commands/schedule_corks/aave_cellar/fees_distributor.rs
hannydevelop: done! added to utils

steward/src/commands/schedule_corks/aave_cellar/fees_distributor.rs
hannydevelop: done! made a `submit_cork` fn in utils.

steward/src/commands/schedule_corks/aave_cellar/fees_distributor.rs
hannydevelop: done

steward/src/commands/schedule_corks/aave_cellar/trust.rs
hannydevelop: done

steward/src/commands/schedule_corks/aave_cellar/fees_distributor.rs
```diff
@@ -0,0 +1,59 @@
+use crate::{application::APP, prelude::*, utils::SubmitCork};
+use abscissa_core::{clap::Parser, Command, Runnable};
+use ethers::abi::AbiEncode;
+use steward_abi::aave_v2_stablecoin::*;
+
+/// Fees Distributor subcommand
+#[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Set fees distributor of target Cellar.\n This command will set the fees distributor of a Cellar. This is a validator only command and can only be run by validators. It sets the fees distributor of a Cellar based on the height specified by \n the validators. Therefore, it'll execute the function when the chain reaches that height. This command also takes the fee distributor's address"
```
EricBolten: It may be worth mentioning in this description that the fees distributor is a Cosmos address that the contract will bridge funds to.

steward/src/commands/schedule_corks/aave_cellar/fees_distributor.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// Block height to schedule cork.
+    #[clap(short = 'b', long)]
+    height: u64,
+}
+
+impl Runnable for FeesDistributorCmd {
+    fn run(&self) {
+        abscissa_tokio::run_with_actix(&APP, async {
+            let mut address = self.new_fees_distributor.as_bytes().to_vec();
```
EricBolten: This would be taking the input string as bytes rather than the actual address bytes. `as_bytes` needs to be called on an `Address` type from `deep_space`, which means converting the string first.

steward/src/utils.rs
```diff
@@ -88,3 +99,49 @@ pub async fn get_eth_provider() -> Result<Provider<Http>, Error> {
 pub fn sp_call_error(message: String) -> Error {
     ErrorKind::SPCallError.context(message).into()
 }
+
+pub struct SubmitCork {
```
EricBolten: Rather than creating another type and implementing a function on it, we could simply define a `submit_scheduled_cork` function that takes a `Cork` and a height as arguments.

EricBolten (commented):
Some suggestions and a fix.

EricBolten:
The command structure was updated during the Aave ABI update, where the subcommand is now the cellar shortname, and a slightly different utility function for sending corks was merged. Given the conflicts, we'll need to update this PR by moving the commands to new filenames and updating them to call the same functions as the existing commands.

EricBolten:
Just a ping @hannydevelop, let's prioritize getting this finished so we can include all the scheduled cork commands in the next point release.

hannydevelop:
@EricBolten all done!

steward/src/utils.rs
```diff
@@ -88,3 +98,31 @@ pub async fn get_eth_provider() -> Result<Provider<Http>, Error> {
 pub fn sp_call_error(message: String) -> Error {
     ErrorKind::SPCallError.context(message).into()
 }
+
+pub async fn submit_schedule_cork(cork: Cork, height: u64) -> Result<TxResponse, CosmosGrpcError> {
```
EricBolten: This function is no longer necessary since it's been replaced by `cork::schedule_cork`.

steward/src/commands/schedule_corks/aave2_clr_s/fees_distributor.rs
```diff
@@ -0,0 +1,61 @@
+use crate::{application::APP, cellars, cork, prelude::*};
+use abscissa_core::{clap::Parser, Command, Runnable};
+use deep_space::Address;
+use ethers::abi::AbiEncode;
+use steward_abi::aave_v2_stablecoin::*;
+
+/// Fees Distributor subcommand
+#[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Set fees distributor of target Cellar.\n This command will set the fees distributor of a Cellar to a Cosmos address that the contract will bridge funds to. This is a validator only command and can only be run by validators. It sets the fees distributor of a Cellar based on the height specified by \n the validators. Therefore, it'll execute the function when the chain reaches that height. This command also takes the fee distributor's address"
```
EricBolten: Nit: there's some redundancy in here we don't need in these command descriptions. The scheduled corks are implicitly validator-only, and we really only need to mention arguments if there's something to discuss about them, the help messages will show them and their comments. Could be simplified to something like: "Set fees distributor Cosmos address of target cellar when chain reaches specified height." Similar note for the rest of the commands.

EricBolten (commented):
A couple small comments.

steward/src/utils.rs
```diff
@@ -16,6 +16,8 @@ use std::{convert::TryFrom, time::Duration};
 use tonic::transport::Channel;
 
 pub const TIMEOUT: Duration = Duration::from_secs(60);
+pub const MESSAGE_TIMEOUT: Duration = Duration::from_secs(10);
```
EricBolten: I don't think either `MESSAGE_TIMEOUT` or `CHAIN_PREFIX` are used any more.

EricBolten (commented):
One last bit of lingering code left over that isn't used any more.

.github/workflows/rust.yml
```diff
@@ -76,5 +76,4 @@ jobs:
       - run: rustup component add clippy
       - uses: actions-rs/cargo@v1
         with:
-          command: clippy
-          args: -- -D warnings
```
EricBolten: Clippy failing on warnings is intentional. The issue we should be fixing is possibly ignoring the code-generated proto files, but it's not necessary to address here.

EricBolten (commented):
The code LGTM but this is not the change to clippy we need.

EricBolten (approved):
LGTM

PR #149, opened 2022-05-25
Replaced mnemonic with private key in eth import command

hannydevelop:
From issue raised in channel:

```
gorc/steward import key commands are different in that gorc takes the key string while steward takes a mnemonic. Chorus recovers the keys to an ethereal keystore directory at startup for security purposes so they need to be able to load the key.
```


EricBolten:
Commenting here to repeat discussion from other channels, but rather than replacing the ability to import from a mnemonic, we should probably support both.

hannydevelop:
> Commenting here to repeat discussion from other channels, but rather than replacing the ability to import from a mnemonic, we should probably support both.

done

steward/src/commands/keys/eth/import.rs
```diff
@@ -1,13 +1,13 @@
...
 use abscissa_core::{clap::Parser, Application, Command, Runnable};
-use k256::pkcs8::ToPrivateKey;
+use k256::{pkcs8::ToPrivateKey, SecretKey};
 use signatory::FsKeyStore;
 use std::path;
 
 #[derive(Command, Debug, Default, Parser)]
 #[clap(
-    long_about = "DESCRIPTION \n\n Import an external Eth key.\n This command will recover a Eth key, storing it in the keystore. \n It takes a keyname and bip39-mnemonic."
+    long_about = "DESCRIPTION \n\n Import an external Eth key.\n This command will recover a Eth key, storing it in the keystore. \n It takes a keyname and Eth private key."
```
EricBolten: This should be updated to specify it takes either a private key or a mnemonic, and they they are mutually exclusive.

steward/src/commands/keys/eth/import.rs
```diff
@@ -38,39 +41,63 @@ impl Runnable for ImportEthKeyCmd {
...
-                key.to_pkcs8_der()
-                    .expect("Could not PKCS8 encod private key")
-            }
-            Err(_) => {
-                let key = rpassword::read_password_from_tty(Some("> Enter your private-key:\n"))
-                    .expect("Could not read private-key");
+        if self.key.is_some() {
+            let key = match self.key.clone() {
+                Some(private_key) => private_key,
+                None => rpassword::read_password_from_tty(Some("> Enter your private-key:\n"))
```
EricBolten: This condition will never be hit, because you're checking `is_some` before ever running the `match`. We would either need two different import commands, or some kind of configuration inside this command, to handle both command-line provided keys or keys obtained over a TTY.

steward/src/commands/keys/eth/import.rs
```diff
@@ -38,39 +41,63 @@ impl Runnable for ImportEthKeyCmd {
...
+                        .parse::<bip32::DerivationPath>()
+                        .expect("Could not parse derivation path");
+
+                    let key =
+                        bip32::XPrv::derive_from_path(seed, &path).expect("Could not derive key");
+                    let key = k256::SecretKey::from(key.private_key());
+                    key.to_pkcs8_der()
+                        .expect("Could not PKCS8 encod private key")
+                }
+                Err(_) => {
```
EricBolten: Looking at it, it seems like this section was intended to read the private key as a fallback option? Also the above version of the private key creation seems to be using `clarity` and is duplicated from gravity bridge, whereas the version here uses `ethers`. There's no need for both, the caller should be specific about whether they're entering a mnemonic or a private key.

steward/tests/acceptance.rs
```diff
@@ -315,7 +315,7 @@ fn eth_keys_import() -> io::Result<()> {
             "eth",
             "import",
             "mykey",
-            "movie tumble tape seven tool session relax youth pet situate bone leave ordinary oxygen silly picture thing fortune genuine attend clerk super seven cement",
+            "21c055b3ac244b34a3e7e7567558e0ee77e51ede5711088da4ec5c7654a6863c",
```
EricBolten: If we're supporting both types we'll need to test both types.

EricBolten (commented):
Some suggestions.

steward/src/commands/keys/eth/import.rs
hannydevelop: I think the best approach is to implement two import commands

EricBolten:
Was reminded of this issue talking with Chorus One today @hannydevelop, let me know when you've had time to respond to the PR feedback here.

steward/src/commands/keys/eth/import_from_key.rs
```diff
@@ -0,0 +1,61 @@
...
+// - [name] required; keyname
+// - (private-key) optional; when absent the user will be prompted to enter it
+impl Runnable for ImportFromKeyCmd {
+    fn run(&self) {
+        let config = APP.config();
+        let keystore = path::Path::new(&config.keystore);
+        let keystore = FsKeyStore::create_or_open(keystore).expect("Could not open keystore");
+
+        let name = self.name.parse().expect("Could not parse name");
+        if let Ok(_info) = keystore.info(&name) {
```
EricBolten: For unused variable captures I believe you can just use `_`.

steward/src/commands/keys/eth/import_from_key.rs
```diff
@@ -0,0 +1,61 @@
+use super::show::ShowKeyCmd;
+use crate::application::APP;
+use abscissa_core::{clap::Parser, Application, Command, Runnable};
+use k256::pkcs8::ToPrivateKey;
+use signatory::FsKeyStore;
+use std::path;
+
+#[derive(Command, Debug, Default, Parser)]
+#[clap(
+    long_about = "DESCRIPTION \n\n Import an external Eth key via private key.\n This command will recover a Eth key, storing it in the keystore. \n It takes a keyname and private key."
```
EricBolten: We may want to specify the expected format of the private key, i.e. the "0x0000..." format and not a path to a private key file or something.

steward/src/commands/keys/eth/import_from_key.rs
```diff
@@ -0,0 +1,61 @@
...
+pub struct ImportFromKeyCmd {
+    /// Eth keyname.
+    pub name: String,
+
+    /// Overwrite key with the same name in the keystore when set to true. Takes a Boolean.
+    #[clap(short, long)]
+    pub overwrite: bool,
+
+    /// private-key optional. When absent you'll be prompted to enter it.
+    pub key: Option<String>,
```
EricBolten, line 22: Should also specify formatting details here.

steward/src/commands/keys/eth/import_from_mnemonic.rs
```diff
@@ -44,33 +44,20 @@ impl Runnable for ImportEthKeyCmd {
...
-        };
-
-        keystore.store(&name, &key).expect("Could not store key");
+            let key = bip32::XPrv::derive_from_path(seed, &path).expect("Could not derive key");
+            let key = k256::SecretKey::from(key.private_key());
+            let key = key
+                .to_pkcs8_der()
+                .expect("Could not PKCS8 encod private key");
+            keystore.store(&name, &key).expect("Could not store key");
+        }
```
EricBolten, line 62: We should error out if we don't get the `Ok` back from this.

steward/tests/acceptance.rs
EricBolten: Can we check if this supports private keys with the `0x` prefix, or does it require them to be in this format? If it doesn't work when using the prefix, we should support it in the command and strip it out if we have to.

steward/tests/acceptance.rs
```diff
@@ -313,7 +313,54 @@ fn eth_keys_import() -> io::Result<()> {
...
+        keystore: keystore_dir_string.clone(),
+        ..Default::default()
+    };
+
+    let config_file_path = key_dir.path().join("config.toml");
+    let config_string = toml::to_string(&config).expect("could not write config to TOML string");
+    fs::write(config_file_path.clone(), config_string).expect("could not write config file");
+
+    let cmd = runner
+        .args(&[
```
EricBolten, line 358: Do we have any tests that make sure the "overwrite" option works?

EricBolten (commented):
Some comments.

steward/src/commands/keys/eth/import_from_mnemonic.rs
hannydevelop: Scenarios where the key can't be stored and others will throw an error. Also, since it won't be possible to return the error to the user, is it worth it? For instance, from the docs and other places I've seen this used, I'll have to use a panic, println, error, or other macros, with a string of message. 

https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html

I used the error macro for this though, let me know what you think.

steward/tests/acceptance.rs
hannydevelop: No, but I have personally tried it. Do you want me to write tests for all the tests that have the  "overwrite" option? Almost all of the keys commands have this option

steward/src/commands/keys/eth/import_from_key.rs
hannydevelop: Done!

steward/src/commands/keys/eth/import_from_key.rs
hannydevelop: Done!

steward/src/commands/keys/eth/import_from_key.rs
hannydevelop: true, Done!

cbrit:
Closing because of staleness

PR #151, opened 2022-07-07
Add Seven Seas Cert

josephterrigno:


EricBolten (approved):
LGTM

For posterity, Seven Seas will be responsible for acting as the strategy provider for the Aave cellar at launch rather than Peggy JV, and this PR reflects that fact by using their CA as default for this initial version.

PR #153, opened 2022-07-13
Update relayer to use gas_multiplier

EricBolten:
`cargo update` for `gravity_bridge` and put `gas_multiplier` into the Ethereum config to pass onto the orchestrator start command.


cbrit (approved):

EricBolten:
I'm not sure why clippy is failing on this PR, it's complaining about a section of code that wasn't touched here, and running clippy with the same arguments on my local branch doesn't throw any warnings, and the integration tests pass. I don't think this is a blocker to merging this code.

PR #156, opened 2022-09-08
Cellar.sol contract support

### LINKED ISSUE #154: Integrate ERC4626 Cellar.sol contract with steward
https://github.com/PeggyJV/cellar-contracts/blob/main/src/base/Cellar.sol

cbrit:
Closes #154


josephterrigno (approved):

integration_tests/cork_test.go
```diff
@@ -22,292 +20,456 @@ import (
...
+// AaveV2Stablecoin rebalance test values used in multiple methods
+const T_AAVE_DAI = "0x6B175474E89094C44Da98b954EedeAC495271d0F"
+const T_AAVE_POOL = "0xE340B25fE32B1011616bb8EC495A4d503e322177"
+const T_AAVE_USDC = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"
+
+// 1*10^18
+const T_AAVE_MIN_ASSETS_OUT = "1000000000000000000"
+
+func (s *IntegrationTestSuite) TestAaveV2Stablecoin() {
+	// s.Run("Submit rebalance request to MockAaveV2Stablecoin", func() {
```
EricBolten: Both of the integration tests are fully commented out. Are they broken?

integration_tests/cork_test.go
```diff
@@ -22,292 +20,456 @@ import (
...
-			res, err := queryClient.QueryCommitPeriod(context.Background(), &corkTypes.QueryCommitPeriodRequest{})
-			if err != nil {
-				return false
-			}
-			if res.VotePeriodStart != res.CurrentHeight {
-				if res.CurrentHeight%10 == 0 {
-					s.T().Logf("current height: %d, period end: %d", res.CurrentHeight, res.VotePeriodEnd)
-				}
-				return false
+func (s *IntegrationTestSuite) waitForEndOfVotePeriod() {
```
EricBolten: Isn't this function exactly the same as `waitForVotePeriod` with a different log string?

integration_tests/cork_test.go
```diff
@@ -22,292 +20,456 @@ import (
...
-			s.Require().Equal(len(result), 1)
+		return false
+	}, 1*time.Minute, 5*time.Second, "cellar event never seen")
+	time.Sleep(time.Duration(2000000000))
+}
 
-			log := result[0]
-			gravity_abi, err := GravityMetaData.GetAbi()
-			s.Require().NoError(err)
+func (s *IntegrationTestSuite) waitForLogicCallEvent(topic common.Hash, invalidationScope []byte, invalidationNonce int) {
```
EricBolten: Was this helper meant to be used? It looks like the ethereum querying is still done manually in each test.

proto/cellar.proto
```diff
@@ -0,0 +1,230 @@
...
+
+syntax = "proto3";
+package steward.v2;
+
+option go_package = "/steward_proto";
+
+/*
+* Represents a function call to a cellar that implements Cellar.sol
+*/
+message Cellar {
```
EricBolten: Given that we have a Cellar V2 approaching in the near term, should we version this?

proto/cellar.proto
```diff
@@ -0,0 +1,230 @@
...
+        */
+        Rebalance rebalance = 5;
+        // Represents function `setStrategistPayoutAddress(address payout)`
+        SetStrategistPayoutAddress set_strategist_payout_address = 6;
+        // Represents function `setWithdrawType(WithdrawType newWithdrawType)`
+        SetWithdrawType set_withdraw_type = 7;
+        // Represents function `swapPositions(uint256 index1, uint256 index2)`
+        SwapPositions swap_positions = 8;
+        // Represents function `setDepositLimit()`
+        SetDepositLimit set_deposit_limit = 9;
```
EricBolten: Was it decided that `SetDepositLimit` and `SetLiquidityLimit` were SP-available functions? The Notion doc seems to indicate they would have been moved to governance.

proto/cellar.proto
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /*
+    * Insert a trusted position to the list of positions used by the cellar at a given index.
+    *
+    * Represents function `addPosition(uint256 index, address position)`
+    */
+    message AddPosition {
+        // Index at which to add the position
+        uint64 index = 1;
```
EricBolten: Are `uint64` types throughout our definitions here sufficient to represent what will be stored in the `uint256` values in the contract on a case-by-case basis?

proto/cellar.proto
```diff
@@ -0,0 +1,230 @@
...
+    * Move assets between positions. To move assets from/to this cellar's holdings, specify
+    the address of this cellar as the `fromPosition`/`toPosition`.
+    *
+    * Represents function `rebalance(address fromPosition, address toPosition,
+    *  uint256 assetsFrom, SwapRouter.Exchange exchange, bytes calldata params)
+    */
+    message Rebalance {
+        string from_position = 1;
+        string to_position = 2;
+        string assets_from = 3;
```
EricBolten: I suppose here is one example of where we know we need the full `uint256` space and must convert from a string then?

steward/src/utils.rs
```diff
@@ -88,3 +88,10 @@ pub async fn get_eth_provider() -> Result<Provider<Http>, Error> {
 pub fn sp_call_error(message: String) -> Error {
     ErrorKind::SPCallError.context(message).into()
 }
+
+pub fn sp_call_parse_address(address: String) -> Result<H160, Error> {
+    match address.parse::<H160>() {
```
EricBolten, line 93: I've seen this a bunch, why do we use the `H160` type for address rather than the ethers `Address` type with more support functions? Does the parse for `Address` contain further validation that `H160` might not entail?

steward/src/cellars/cellar.rs
```diff
@@ -0,0 +1,228 @@
...
+                    ErrorKind::SPCallError.context("swap params cannot be empty")
+                })?)?;
+
+            info!("encoded: {:?}", hex::encode(&swap_params));
+
+            let call = RebalanceCall {
+                from_position: sp_call_parse_address(params.from_position)?,
+                to_position: sp_call_parse_address(params.to_position)?,
+                assets_from: string_to_u256(params.assets_from)?,
+                exchange: (params.exchange - 1) as u8,
```
EricBolten: Curious about the `-1` here...wouldn't that make the enum values incorrect? It looks like here:

https://github.com/PeggyJV/cellar-contracts/blob/03ef9c1fc37ed5630964ad05ca3a62b4cbcba358/src/modules/swap-router/SwapRouter.sol#L24-L26

That the enum doesn't have an "unspecified" for the zero value, if those two are assigned to zero and one respectively. Why did we add an extra type and do this subtraction?

steward/src/cellars/cellar.rs
```diff
@@ -0,0 +1,228 @@
...
+            Ok(CellarCalls::SetStrategistPayoutAddress(call).encode())
+        }
+        Function::SetWithdrawType(params) => {
+            log_cellar_call(
+                CELLAR_NAME,
+                &SetWithdrawTypeCall::function_name(),
+                &cellar_id,
+            );
+            let call = SetWithdrawTypeCall {
+                new_withdraw_type: params.new_withdraw_type as u8,
```
EricBolten: We're setting an enum in the same fashion here as exhcange type above, except there's no `-1` even though the two enums in the contract have two entries and our enums have three entries respectively.

EricBolten (commented):
Some comments and questions. The other key bit we'll want to add for the new cellar though is the scheduled cork functions for the governance-gated calls. Even though we'll be adding the ability to do this directly via governance in the next cork version, we still want a utility to manually call them.

Makefile
```diff
@@ -15,27 +15,20 @@ e2e_build_images: e2e_clean_slate
...
-		2>/dev/null \
-		|| true
-	@docker network prune --force 1>/dev/null 2>/dev/null || true
-	@cd integration_tests && go test -c
-
-e2e_cork_test: e2e_clean_slate
-	@E2E_SKIP_CLEANUP=true integration_tests/integration_tests.test -test.failfast -test.v -test.run IntegrationTestSuite -testify.m TestCork || make -s fail
+	@./clean_slate.sh
+
+e2e_cork_test: e2e_aave_v2_stablecoin_test e2e_vault_cellar_test
```
EricBolten: If we're going to do this, we should update the integration test action to call the two different integration tests separately so they could be parallelized.

proto/cellar.proto
cbrit: I asked the contract guys about this. I figured it could be Cellar/CellarV2 but open to suggestions

proto/cellar.proto
cbrit: Yes I was requested to add them. CC @josephterrigno

proto/cellar.proto
cbrit: Right

steward/src/utils.rs
cbrit: Just habit. H160 was used early on in `cellar_rebalancer` development and I'm stuck in my ways. Will make the switch.

steward/src/cellars/cellar.rs
cbrit: Protobuf forces you to have a zero value for unspecified so this is the necessary conversion

integration_tests/cork_test.go
cbrit: Oops

integration_tests/cork_test.go
cbrit: It is, renamed to make it clearer. This is specifically for the Gravity contract, the manual one is for events in the mock cellar.

proto/cellar.proto
cbrit: I can see it going either way. Should we just always default to string if it's uint256?

steward/src/cellars/cellar.rs
cbrit: Oof good catch. It needs the -1

proto/cellar.proto
EricBolten: Yeah, just figured maybe `CellarV1` or something would be more appropriate so that there isn't an "unversioned" cellar name, but not critical.

proto/cellar.proto
EricBolten: Gotcha, just wanted to confirm.

proto/cellar.proto
EricBolten: The Aave cellar has a mix, but this is an opportunity to change that. It really depends on the expected values for the argument -- leaving it as a string puts no restrictions on it but takes up more wire space, though relatively negligible in this context. If the contract team can confirm the value will always exist within the bounds of the max size of the input (e.g. for `uint64`) then it should be OK, but in truth I think it might be safer to just universally handle all `uint256` the same.

steward/src/cellars/cellar.rs
EricBolten: We may want to add some comments to this effect to make it clear why the code is doing this.

proto/cellar.proto
cbrit: Let's go with the latter

steward/src/commands/schedule_corks/cellar_v1/set_fees_distributor.rs
```diff
@@ -0,0 +1,58 @@
...
+
+/// Fees Distributor subcommand
+#[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION\n\nCalls setFeesDistributor() on the target cellar contract at the specified block height.\nFor more information see https://github.com/PeggyJV/cellar-contracts/blob/main/src/base/Cellar.sol"
+)]
+pub struct SetFeesDistributorCmd {
+    #[clap(short = 'n', long)]
+    /// Fee distributor's address
+    new_fees_distributor: Address,
```
EricBolten, line 15: This is the ethers `Address` type, and the bytes have to come from a Cosmos address.

steward/src/commands/schedule_corks/cellar_v1/set_owner.rs
```diff
@@ -0,0 +1,50 @@
...
+use steward_abi::cellar::*;
+
+/// Fees Distributor subcommand
+#[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION\n\nCalls setOwner() on the target cellar contract at the specified block height.\nFor more information see https://github.com/PeggyJV/cellar-contracts/blob/main/src/base/Cellar.sol"
+)]
+pub struct SetOwnerCmd {
+    #[clap(short = 'n', long)]
+    /// Fee distributor's address
```
EricBolten: This comment was from another command.

proto/cellar_v1.proto
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /*
+    * Insert a trusted position to the list of positions used by the cellar at a given index.
+    *
+    * Represents function `addPosition(uint256 index, address position)`
+    */
+    message AddPosition {
+        // Index at which to add the position
+        uint64 index = 1;
```
EricBolten: I think you said in the earlier review you wanted to adjust these to strings but they are still typed as `uint64`.

steward/src/cellars/cellar_v1.rs
```diff
@@ -0,0 +1,228 @@
...
+                    ErrorKind::SPCallError.context("swap params cannot be empty")
+                })?)?;
+
+            info!("encoded: {:?}", hex::encode(&swap_params));
+
+            let call = RebalanceCall {
+                from_position: sp_call_parse_address(params.from_position)?,
+                to_position: sp_call_parse_address(params.to_position)?,
+                assets_from: string_to_u256(params.assets_from)?,
+                exchange: (params.exchange - 1) as u8,
```
EricBolten, line 84: The other follow-up from earlier is adding comments explaining this.

steward/src/cellars/cellar_v1.rs
```diff
@@ -0,0 +1,228 @@
...
+            Ok(CellarCalls::SetStrategistPayoutAddress(call).encode())
+        }
+        Function::SetWithdrawType(params) => {
+            log_cellar_call(
+                CELLAR_NAME,
+                &SetWithdrawTypeCall::function_name(),
+                &cellar_id,
+            );
+            let call = SetWithdrawTypeCall {
+                new_withdraw_type: (params.new_withdraw_type - 1) as u8,
```
EricBolten, line 112: Here as well.

EricBolten (commented):
A few comments and follow-ups.

steward/src/cellars/cellar_v1.rs
```diff
@@ -28,7 +28,7 @@ pub fn get_encoded_call(function: Function, cellar_id: String) -> Result<Vec<u8>
         Function::AddPosition(params) => {
             log_cellar_call(CELLAR_NAME, &AddPositionCall::function_name(), &cellar_id);
             let call = AddPositionCall {
-                index: params.index.into(),
+                index: utils::string_to_u256(params.index)?,
```
EricBolten: Nit: we use both `utils::string_to_u256` and a direct import of `string_to_u256` throughout the file.

integration_tests/ethereum/contracts/MockCellar.json
```diff
@@ -67,5 +105,25 @@
...
+    },
+    {
+        "inputs": [
+            {
+                "internalType": "address",
+                "name": "newOwner",
+                "type": "address"
+            }
+        ],
+        "name": "transferOwnership",
```
EricBolten: These functions come from `Ownable.sol`, but the contract is using `Owned.sol`. If we need to add them to the mock contract, they'd have to be the `setOwner` version with its associated events.

EricBolten (commented):
Small comments.

EricBolten (approved):
LGTM

PR #168, opened 2022-11-14
Compare binary and release versions on release

cbrit:
A naive version check, comparing the `Cargo.toml` package version with the release tag to prevent releasing a binary with a version mismatch. It's a little brittle as the `grep` command used assume the `[package]` section is at the top of the TOML and thus will be the first match of "^version = ...". My bash-fu is weak, so please recommend a more reliable command if needed.

If/when this is merged I'll do the same for the orchestrator.

### DIFF
--- a/check_version.sh
+++ b/check_version.sh
@@ -4,7 +4,7 @@ set -e
 
 # this is a little fragile because it assumes the [package] section will be at the
 # top of the Cargo.toml
-BINARY_VERSION_STRING=$(cat steward/Cargo.toml | grep "^(version = \")([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+[0-9A-Za-z-]+)?" | head -1)
+BINARY_VERSION_STRING=$(cat steward/Cargo.toml | grep "^\(version = \"\)\([0-9]\+\)\.\([0-9]\+\)\.\([0-9]\+\)\(?:-\([0-9A-Za-z-]\+\(?:\.[0-9A-Za-z-]\+\)*\)\)\?\(?:\+[0-9A-Za-z-]\+\)\?" | head -1)
 # above output is 'version = "x.x.x"' so we trim the leading and trailing characters
 BINARY_VERSION=${BINARY_VERSION_STRING:11:-1}
 # trim the leading 'v' from the tag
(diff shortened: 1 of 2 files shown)


check_version.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -e
+
+# this is a little fragile because it assumes the [package] section will be at the
+# top of the Cargo.toml
+BINARY_VERSION_STRING=$(cat steward/Cargo.toml | grep "^version.*" | head -1)
+# above output is 'version = "x.x.x"' so we trim the leading and trailing characters
```
EricBolten, line 8: This would likely be more robust if we use a standard semver regex with a capture group for the version string. Not sure if this accounts for trailing elements, etc.

EricBolten (commented):
A small suggestion.

check_version.sh
cbrit: I can do that; it will still capture any other toml tables (sections) that have a version line, such as `[dependencies.abscissa_core]` in the steward toml. Are we concerned with that right now? We typically keep the package section at the top.

EricBolten: It's probably fine, considering this is a workflow for this specific repo and we shouldn't have expectations of that shifting out from under us. In the more paranoid case if you just wanted to stick to regex you could look for the `[package]` header as part of your pattern matching.

check_version.sh
```diff
@@ -0,0 +1,19 @@
...
+
+set -e
+
+# this is a little fragile because it assumes the [package] section will be at the
+# top of the Cargo.toml
+BINARY_VERSION_STRING=$(cat steward/Cargo.toml | grep "^\(version = \"\)\([0-9]\+\)\.\([0-9]\+\)\.\([0-9]\+\)\(?:-\([0-9A-Za-z-]\+\(?:\.[0-9A-Za-z-]\+\)*\)\)\?\(?:\+[0-9A-Za-z-]\+\)\?" | head -1)
+# above output is 'version = "x.x.x"' so we trim the leading and trailing characters
+BINARY_VERSION=${BINARY_VERSION_STRING:11:-1}
+# trim the leading 'v' from the tag
+RELEASE_VERSION=${GITHUB_REF_NAME: 1}
```
EricBolten, line 11: Can there be a space here in this substring syntax? I assume it was tested but seems odd.

EricBolten: Hrmm, example of parameter expansion seem to indicate that this will work fine.

EricBolten (approved):
LGTM

PR #171, opened 2022-12-05
CI test runners

EricBolten:
Testing out adjusting runners.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -181,5 +181,5 @@ jobs:
           E2E_SKIP_CLEANUP: true
       - name: Clean up docker images
         run: ./clean_slate.sh
-      - name: Print file permissions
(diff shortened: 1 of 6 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -112,13 +112,17 @@ jobs:
...
     needs: [ steward-build, hardhat-build ]
     strategy:
       fail-fast: false
       matrix:
         test_type: [ "AaveV2Stablecoin", "CellarV1" ]
     steps:
+      - name: Set up Go 1.16
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.16
```
cbrit: Nit: We saw earlier that at least one of our repos is 1.15. Not sure if we care if this is canonical in CI.

cbrit (approved):

PR #172, opened 2022-12-09
Edit cosmos gas_adjustment value recommendation

cbrit:
Edit cosmos gas_adjustment value recommendation. Some orchestrators are failing to submit ethereum events because their adjustment is set to 1.0 and `gasWanted` is set just under `gasRequired` in their tx.

### DIFF
--- a/docs/01-Configuration.md
+++ b/docs/01-Configuration.md
@@ -39,11 +39,11 @@ Configuration related interactions with the Cosmos chain in question
 
 Type: float
 
-Multiplied by estimated gas fee per transaction. Currently Sommelier fee requirements are 0.0 so this can be left as default.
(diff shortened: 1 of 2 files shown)


docs/01-Configuration.md
```diff
@@ -302,7 +302,7 @@ This example will not work as is, you'll need to supply your own values.
 keystore = "/some/path/keystore"
 
 [cosmos]
-gas_adjustment = 1.0
+gas_adjustment = 1.1
```
EricBolten, line 305: If we're recommending this for the configuration, should we just set the default to 1.1?

EricBolten (commented):
A question.

EricBolten (approved):
LGTM

PR #176, opened 2023-01-13
Update gravity to 2.0.4 for 2.x line

EricBolten:


cbrit (commented):
Could you add the check-version step to the workflow for this branch?

cbrit (approved):

PR #177, opened 2023-01-16
Steward 2.x line support for Cellar V2 architecture

cbrit:


integration_tests/ethereum/contracts/MockAdaptor.json
```diff
@@ -0,0 +1 @@
...untToBorrow","type":"uint256"}],"name":"borrowFromAave","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"contract ERC20","name":"assetOut","type":"address"},{"internalType":"enum Cellar.Exchange","name":"exchange","type":"uint8"},{"internalType":"bytes","name":"params","type":"bytes"},{"internalType":"uint64","name":"slippage","type":"uint64"}],"name":"claimCompAndSwap","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[],"name":"owner","outputs":[{"internalType":"address","name":"","type":"address"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"newOwner","type":"address"}],"name":"setOwner","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"contract ERC20","name":"tokenIn","type":"address"},{"internalType":"contract ERC20","name":"tokenToRepay","type":"address"},{"internalType":"uint256","name":"amountIn","type":"uint256"},{"internalType":"enum Cellar.Exchange","name":"exchange","type":"uint8"},{"internalType":"bytes","name":"params","type":"bytes"}],"name":"swapAndRepay","outputs":[],"stateMutability":"nonpayable","type":"function"}]
```
EricBolten: Nit: storing this as prettified would make it easier to read.

integration_tests/ethereum/contracts/MockCellar.json
```diff
@@ -1 +1 @@
...
...ata","type":"bytes[]"}],"internalType":"struct Cellar.AdaptorCall[]","name":"data","type":"tuple[]"}],"name":"callOnAdaptor","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[],"name":"owner","outputs":[{"internalType":"address","name":"","type":"address"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"fromPosition","type":"address"},{"internalType":"address","name":"toPosition","type":"address"},{"internalType":"uint256","name":"assetsFrom","type":"uint256"},{"internalType":"enum Cellar.Exchange","name":"exchange","type":"uint8"},{"internalType":"bytes","name":"params","type":"bytes"}],"name":"rebalance","outputs":[{"internalType":"uint256","name":"assetsTo","type":"uint256"}],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"address","name":"newOwner","type":"address"}],"name":"setOwner","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"address","name":"position","type":"address"},{"internalType":"enum Cellar.PositionType","name":"positionType","type":"uint8"}],"name":"trustPosition","outputs":[],"stateMutability":"nonpayable","type":"function"}]
```
EricBolten: Same nit here as above.

steward_abi/CellarV2.json
```diff
@@ -0,0 +1,1960 @@
+[
```
EricBolten, line 1: Noting it as a comment here rather than in `CellarV1.json` since github won't even render it: why is the CellarV1 JSON 20x larger than CellarV2?

proto/adaptors/aave/a_token.proto
```diff
@@ -0,0 +1,59 @@
...
+
+import "adaptors/base.proto";
+
+// Represents call data for the Aave AToken adaptor, used to manage lending positions on Aave
+message AaveATokenAdaptor {
+    oneof function {
+        /***** BASE ADAPTOR FUNCTIONS *****/
+
+        // Represents function `swap(ERC20 assetIn, ERC20 assetOut, uint256 amountIn, SwapRouter.Exchange exchange, bytes memory params)`
+        Swap swap = 3;
```
EricBolten: This proto definition is a tad confusing using out-of-order IDs.

proto/cellar_v2.proto
```diff
@@ -0,0 +1,167 @@
...
+        // Whether to switch positions in the debt array, or the credit array.
+        bool in_debt_array = 3;
+    }
+
+    /*
+    * Set the per-wallet deposit limit. Uses the same decimals as the current asset.
+    *
+    * Represents function `setDepositLimit()`
+    */
+    message SetDepositLimit {
```
EricBolten: Where is `setDepositLimit` in the cellar contract?

For reference am looking at https://github.com/PeggyJV/cellar-contracts/blob/main/src/base/Cellar.sol

proto/cellar_v2.proto
```diff
@@ -0,0 +1,167 @@
...
+        string new_limit = 1;
+    }
+
+    /*
+    * Set the maximum liquidity that cellar can manage. Uses the same decimals as the current asset.
+    *
+    * Represents function `setLiquidityLimit()`
+    */
+    message SetLiquidityLimit {
```
EricBolten: Same here, I don't see this function.

proto/cellar_v2.proto
```diff
@@ -0,0 +1,167 @@
...
+    *
+    * Represents function `addPosition(uint32 index, uint32 positionId, bytes configurationData, bool inDebtArray)`
+    */
+    message AddPosition {
+        // Index at which to add the position
+        uint32 index = 1;
+        // The position's ID in the cellar registry
+        uint32 position_id = 2;
+        // Data used to configure how the position behaves
+        bytes configuration_data = 3;
```
EricBolten, line 53: Is there a format we should be providing to serialize `configuration_data` rather than expecting raw bytes?

steward/src/cellars/cellar_v2.rs
```diff
@@ -0,0 +1,702 @@
...
+        }
+    }
+}
+
+fn get_encoded_adaptor_call(data: Vec<AdaptorCall>) -> Result<Vec<AbiAdaptorCall>, Error> {
+    let mut result: Vec<AbiAdaptorCall> = Vec::new();
+    for d in data {
+        debug!("adaptor call to {}", d.adaptor);
+        let mut calls: Vec<Bytes> = Vec::new();
+        match d.call_data.unwrap() {
```
EricBolten: Are there any checks earlier on to confirm this `Option` is unwrappable? Otherwise don't we run the risk of a panic?

steward/src/cellars/cellar_v2.rs
```diff
@@ -0,0 +1,702 @@
...
+                                min_1: string_to_u256(p.min_1)?,
+                                collect_fees: p.collect_fees,
+                            };
+                            calls.push(
+                                UniswapV3AdaptorCalls::TakeFromPosition(call)
+                                    .encode()
+                                    .into(),
+                            );
+                        }
+                        uniswap_v3_adaptor::Function::Swap(p) => {
```
EricBolten, line 193: Is there a way to create some common functionality for the encoding of the base adaptor functions so we don't have to reimplement `Swap`, `OracleSwap`, and `RevokeApproval` for every serialization?

steward/src/cellars/cellar_v2.rs
```diff
@@ -0,0 +1,702 @@
...
+
+fn get_encoded_adaptor_call(data: Vec<AdaptorCall>) -> Result<Vec<AbiAdaptorCall>, Error> {
+    let mut result: Vec<AbiAdaptorCall> = Vec::new();
+    for d in data {
+        debug!("adaptor call to {}", d.adaptor);
+        let mut calls: Vec<Bytes> = Vec::new();
+        match d.call_data.unwrap() {
+            CallData::UniswapV3Calls(params) => {
+                for c in params.calls {
+                    match c.function.expect("adaptor function was empty") {
```
EricBolten: Similarly here, do we want to `expect` and panic if a value is missing? Don't protos have to have default zero values in the case of missing fields since "required" isn't really a concept in proto v3? I think in other places we use the `ok_or_else` method.

steward/src/cellars/cellar_v2.rs
```diff
@@ -0,0 +1,702 @@
...
+                        uniswap_v3_adaptor::Function::OpenPosition(p) => {
+                            let call = steward_abi::uniswap_v3_adaptor::OpenPositionCall {
+                                token_0: sp_call_parse_address(p.token_0)?,
+                                token_1: sp_call_parse_address(p.token_1)?,
+                                pool_fee: p.pool_fee,
```
EricBolten, line 151: I guess the generated ABI assigns stuff like `uint24 poolFee` to a `uint32` and somehow handles conversion on the backend? We're matching the generated ABI so there's nothing we could do, but it's an interesting question.

steward/src/cellars/cellar_v2.rs
```diff
@@ -0,0 +1,702 @@
...
+                                amount_0: string_to_u256(p.amount_0)?,
+                                amount_1: string_to_u256(p.amount_1)?,
+                                min_0: string_to_u256(p.min_0)?,
+                                min_1: string_to_u256(p.min_1)?,
+                                tick_lower: p.tick_lower as i32,
```
EricBolten: If `tickLower` and `tickUpper` are defined as `int24` in the contract, should we not natively use `int32` rather than `uint32` in our proto?

steward/src/cellars/cellar_v2.rs
```diff
@@ -0,0 +1,702 @@
...
+                            let swap_params = encode_swap_params(p.params.ok_or_else(|| {
+                                ErrorKind::SPCallError.context("swap params cannot be empty")
+                            })?)?;
+
+                            debug!("encoded: {:?}", hex::encode(&swap_params));
+                            let call = steward_abi::uniswap_v3_adaptor::SwapCall {
+                                asset_in: sp_call_parse_address(p.asset_in)?,
+                                asset_out: sp_call_parse_address(p.asset_out)?,
+                                amount_in: string_to_u256(p.amount_in)?,
+                                exchange: convert_exchange(p.exchange),
```
EricBolten, line 203: `exchange` comes in as an `i32` despite our always non-negative enum values -- I didn't see it explicitly set anywhere, is this how the proto is interpreting an `enum`?

steward/src/cellars/cellar_v2.rs
```diff
@@ -0,0 +1,702 @@
...
+            };
+
+            Ok(CellarV2Calls::SetRebalanceDeviation(call).encode())
+        }
+    }
+}
+
+/// Encodes the swap params as ABI-encoded bytes to be passed as args to the underlying
+/// swap function
+fn encode_swap_params(params: SwapParams) -> Result<Vec<u8>, Error> {
```
EricBolten: `SwapParams` and `OracleSwapParams` are defined as "common" protos, but we're defining their implementation in the `cellar_v2` file. I believe we duplication `SwapParams` encoding in `cellar_v1`. Should we move to a common location?

steward/src/cellars/cellar_v2.rs
```diff
@@ -0,0 +1,702 @@
...
+                    }
+                }
+            }
+            CallData::CompoundCTokenCalls(params) => {
+                for c in params.calls {
+                    match c.function.expect("adaptor function was empty") {
+                        compound_c_token_adaptor::Function::DepositToCompound(p) => {
+                            let call =
+                                steward_abi::compound_c_token_adaptor::DepositToCompoundCall {
+                                    market: sp_call_parse_address(p.market)?,
```
EricBolten, line 406: It seems like based on the generated ABI `CErc20` for Compound are treated the same as `ERC20` in types of expecting an address. None of our mock calls are using that type though -- it would be great if we could confirm this serializes as expected.

steward/src/cellars/cellar_v2.rs
```diff
@@ -0,0 +1,702 @@
...
+                                    ErrorKind::SPCallError.context("swap params cannot be empty")
+                                })?)?;
+
+                            debug!("encoded: {:?}", hex::encode(&oracle_swap_params));
+                            let call =
+                                steward_abi::compound_c_token_adaptor::ClaimCompAndSwapCall {
+                                    asset_out: sp_call_parse_address(p.asset_out)?,
+                                    // to account for protobuf's requirement that an UNSPECIFIED enum variant be defined
+                                    // as 0, we subtract 1 from the value
+                                    exchange: (p.exchange - 1) as u8,
```
EricBolten: Don't we have `convert_exchange` for this?

steward/src/cellars/cellar_v2.rs
```diff
@@ -0,0 +1,702 @@
...
+                        }
+                    }
+                }
+            }
+            CallData::VestingSimpleCalls(params) => {
+                for c in params.calls {
+                    match c.function.expect("adaptor function was empty") {
+                        vesting_simple_adaptor::Function::DepositToVesting(p) => {
+                            let call = steward_abi::vesting_simple_adaptor::DepositToVestingCall {
+                                vesting_contract: sp_call_parse_address(p.vesting_contract)?,
```
EricBolten, line 506: Similar to `CErc20`, the actual input in the contract here is a `VestingSimple` rather than an `ERC20`...again, the ABI generated code indicates this should work, just wondering if we also want to include a test in the mock contract where these types are imported and used.

steward/src/cellars/cellar_v2.rs
```diff
@@ -0,0 +1,702 @@
...
+                configuration_data: params.configuration_data.into(),
+                in_debt_array: params.in_debt_array,
+            };
+
+            Ok(CellarV2Calls::AddPosition(call).encode())
+        }
+        CallOnAdaptor(params) => {
+            log_cellar_call(CELLAR_NAME, &CallOnAdaptorCall::function_name(), &cellar_id);
+            let call = CallOnAdaptorCall {
+                data: get_encoded_adaptor_call(params.data)?,
```
EricBolten, line 48: Not sure if this is something steward should concern itself with, but is there any upper limit on how much data can be passed into a contract function call? Seems like this could get arbitrarily large.

EricBolten (commented):
Some feedback.

steward_abi/CellarV2.json
cbrit: I think I compiled the CellarV1 ABI myself and it included the bytecode and metadata, whereas V2 I copied from etherscan.

proto/cellar_v2.proto
cbrit: The answer I got is they would just encode it on their end. I actually don't know how technically difficult it would be to do given each adaptor has its own structure.

steward/src/cellars/cellar_v2.rs
cbrit: The way enums work in rust force us into this situation afaict; the only way I could think to minimize the rust-side code is to complicate the proto, such as making BaseCall an `AdaptorCall` option in which they would have to specify which adaptor they are targeting somehow.

steward/src/cellars/cellar_v2.rs
cbrit: we can warn the SPs not to pick the >=2^24 fee pools :D

steward/src/cellars/cellar_v2.rs
cbrit: yeah it's weird

steward/src/cellars/cellar_v2.rs
cbrit: It seems like the ABI calling them the same type is definitive enough to me, but I can add this tomorrow if you disagree.

steward/src/cellars/cellar_v2.rs
cbrit: My thinking was since the ABI defines them as the same underlying type, I didn't want to set a precedent of digging up solidity code to import into the mock contract every time an adaptor does this. But again, I can do it if necessary.

steward/src/cellars/cellar_v2.rs
cbrit: Annoyingly hard to find an answer to. Can only find info about max gas.

steward/src/cellars/cellar_v2.rs
cbrit: tbh is could (maybe should) be an enum in the steward proto since iirc there are only three discrete fee levels on uniswap V3: 0.05%, 0.30%, 1%. Kind of a gross enum. Maybe Steward could just validate the value?

steward/src/cellars/cellar_v2.rs
cbrit: Adding a `BaseCall` in the proto

steward/src/cellars/cellar_v2.rs
EricBolten: I think this kind of depends on how likely it is that that might ever change, since it would mean we'd have to roll out a new steward version to support other fees. This might be a question we throw back to the SPs to ask if it would be better to do it this way. We did have one failing call a while ago because the fee value was set incorrectly.

Cargo.lock
```diff
@@ -4856,6 +4856,7 @@ version = "2.0.0"
 dependencies = [
  "prost 0.7.0",
  "prost-types 0.7.0",
+ "serde",
```
EricBolten, line 4859: Putting this comment into Cargo.lock even though it isn't relevant to this line: this PR needs a version bump for the next 2.x release.

proto/cellar_v2.proto
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    // Represents a call to adaptor an. The cellar must be authorized to call the target adaptor.
+    message AdaptorCall {
+        // Address of the adaptor
+        string adaptor = 1;
+        // The function call data for the adaptor
+        oneof call_data {
+            // Represents base adaptor function calls native to all adaptors
+            BaseAdaptorCalls base_calls = 2;
```
EricBolten: Having made this change, I am realizing there is a potential issue that could make using it weird. It's great that we've avoided code duplication, but I think this structure means that if say an `AdaptorCall` was to be made for Uniswap and it interleaved adaptor and base adaptor calls, it would have to be split up into multiple `AdaptorCall` protos because it's a `oneof`. We used this method of having a `oneof` for call data and then having each adaptor be a repeated message -- should we instead create an `AdaptorCallData` message, have that be a repeated field here and within it contain a `oneof` of singular adaptor calls? This would allow mixing and matching of base adaptor calls and adaptor calls, though with the side effect of it technically being possible to mix different adaptors together (which wouldn't work) if you did it incorrectly. Otherwise, we'd have to move base adaptor calls back into the adaptor to allow them to be included in one `AdaptorCall` (thankfully with your commit structure we could just drop that commit if we chose to do so, but...).

steward/src/commands/schedule_corks/cellar_v1/lift_shutdown.rs
```diff
@@ -1,7 +1,7 @@
 use crate::{application::APP, cellars, cork, prelude::*};
 use abscissa_core::{clap::Parser, Command, Runnable};
 use ethers::abi::AbiEncode;
-use steward_abi::cellar::*;
+use steward_abi::cellar_v1::*;
```
EricBolten, line 4: Tagging this in a schedule command even though it's not relevant to the line: should we include scheduled cork calls for the governance functions of Cellar V2? Without them, we can't execute one of those functions in the intervening time before the upgrade.

EricBolten (commented):
Some follow-up thoughts.

steward/src/commands/schedule_corks/cellar_v1/lift_shutdown.rs
cbrit: Since such a call would require coordination on the part of the validator set to use anyway, perhaps we can just upgrade this 2.x line with it if/when the issue comes up, and then have them upgrade/do the command at the same time?

EricBolten: We can prep this code after cutting this release.

EricBolten (approved):
LGTM

PR #186, opened 2023-02-09
Initial Tenderly simulation mode

### LINKED ISSUE #193: Tenderly simulation mode for strategists

cbrit:
Closes #193


proto/steward.proto
```diff
@@ -41,3 +49,21 @@ message ScheduleResponse {
     // The hex encoded ID of the scheduled cork
     string id = 1;
 }
+
+/*
+ * Represents a simulated function call to a particular Cellar
+ */
+ message SimulateRequest {
+    ScheduleRequest request = 1;
```
EricBolten, line 67: We may want to use a different message type here, unless it makes a bunch more downstream changes necessary -- no block height required to "schedule" a simulated transaction.

proto/steward.proto
```diff
@@ -41,3 +49,21 @@ message ScheduleResponse {
...
+    /// Whether to simply encode and return the contract call data, skipping the Tenderly simulation
+    bool encode_only = 2;
+}
+
+message SimulateResponse {
+    /// The encoded contract call
+    string encoded_call = 1;
```
EricBolten, line 74: Are we returning this as a hex encoded string? Should we send the raw bytes?

steward/src/commands/start.rs
```diff
@@ -10,29 +10,43 @@ use crate::{
...
+pub struct StartCmd {
+    /// Enable simulation mode
+    /// TO-DO: There is a bug with the Override implementation not updating the config properly.
+    /// For now we will just have to explicitly set enable to `true` in the config file.
+    #[clap(long)]
+    pub simulate: bool,
+
+    /// TO-DO: Same problem as above.
+    #[clap(long)]
+    pub tls: bool,
```
EricBolten: With regards to these two options, I don't think we should add flags to the "start" mode (we especially don't want to give the option of turning TLS off!). We can have a separate command called "simulator" or something rather than start so it's quite obvious it's running in a different mode, and handle those types of options there.

steward/src/commands/start.rs
```diff
@@ -76,7 +115,15 @@ impl config::Override<StewardConfig> for StartCmd {
     // Process the given command line options, overriding settings from
     // a configuration file using explicit flags taken from command-line
     // arguments.
-    fn override_config(&self, config: StewardConfig) -> Result<StewardConfig, FrameworkError> {
+    fn override_config(&self, mut config: StewardConfig) -> Result<StewardConfig, FrameworkError> {
```
EricBolten: We shouldn't have to worry about overriding if we split the commands out.

steward/src/simulate.rs
```diff
@@ -0,0 +1,84 @@
...
+
+pub struct SimulateHandler;
+
+#[async_trait]
+impl steward::simulate_contract_call_server::SimulateContractCall for SimulateHandler {
+    async fn simulate(
+        &self,
+        request: Request<SimulateRequest>,
+    ) -> Result<Response<SimulateResponse>, Status> {
+        if !APP.config().simulate.enabled {
```
EricBolten: I'm not sure we need to run what are essentially meant to be compile-time problems every time a call is invoked. Splitting out the modes may help this feel less awkward to remove.

steward/src/simulate.rs
```diff
@@ -0,0 +1,84 @@
...
+        return Err(ErrorKind::Http.context("empty contract call data").into());
+    }
+
+    match request.call_data.unwrap() {
+        AaveV2Stablecoin(call) => {
+            if call.function.is_none() {
+                return Err(ErrorKind::Http.context("empty function data").into());
+            }
+
+            aave_v2_stablecoin::get_call(call.function.unwrap(), request.cellar_id)
```
EricBolten: `get_call` returns the enum value, do you mean this to be a stringified version of `get_encoded_call`? Will `to_string` output hex for a `Vec<u8>`?

steward/src/tenderly.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub save_if_fails: bool,
+    pub simulation_type: String,
+}
+
+pub async fn simulate(cellar_id: String, encoded_call: String) -> Result<String, Status> {
+    let config = APP.config();
+    validate_tenderly_config(&config);
+
+    let body = serde_json::to_string(&SimulateRequest {
+        network_id: "1".to_string(),
```
EricBolten: Will be more future-proof to move this field into the config rather than just hardcoding it here. Plus in theory one might want to, even now, run this sort of test against Goerli.

steward/src/tenderly.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub to: String,
+    pub input: String,
+    pub save: bool,
+    pub save_if_fails: bool,
+    pub simulation_type: String,
+}
+
+pub async fn simulate(cellar_id: String, encoded_call: String) -> Result<String, Status> {
+    let config = APP.config();
+    validate_tenderly_config(&config);
```
EricBolten: Can this be done on startup rather than on every simulate call?

steward/src/tenderly.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    let body = serde_json::to_string(&SimulateRequest {
+        network_id: "1".to_string(),
+        from: GRAVITY_ADDRESS.to_string(),
```
EricBolten: Similarly here for testnet purposes, we may want to make this an adjustable config value rather than hardcoded.

steward/src/tenderly.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    client
+        .execute(request)
+        .await
+        .unwrap()
+        .text()
+        .await
+        .map_err(|e| {
+            error!("simulate error: {}", e);
+            Status::new(Code::Unknown, format!("simulate error: {e}"))
```
EricBolten: Does `reqwest` not attach a code to its error?

EricBolten (commented):
Some thoughts on structure and a few comments.

steward/src/commands/start.rs
cbrit: doesn't work anyway; i reported a bug in the abscissa repo

steward/src/simulate.rs
cbrit: that's the idea, and it seems to have worked. I wasn't sure what the simplest way to get the ethereum-formatted hex was other than to do this.

steward/src/tenderly.rs
cbrit: My thinking was this is the status of steward, not tenderly, that we should return?

proto/steward.proto
cbrit: It was to avoid downstream changes. `get_encoded_call` takes in a `ScheduleRequest`.

proto/steward.proto
cbrit: The goal is to show it in the format you would read it in tenderly or etherscan, the ethereum-formatted hex encoded call string. This is also the form the Tenderly API takes in the data.

cbrit: heh and it's broken. similar to how H160.to_string() doesn't work, it just returns the zero address

cbrit: heh and it's broken. similarly to how H160.to_string() just returns the zero address. Somehow the zero address is considered valid input data to the simulator so I didn't investigate when I saw a green run.

EricBolten:
Pending merge conflict resolution, will re-review this.

steward/src/cellars/aave_v2_stablecoin.rs
```diff
@@ -25,6 +25,13 @@ const CELLAR_NAME: &str = "aave_v2_stablecoin";
 const LOG_PREFIX: &str = "AaveV2StablcoinCellar";
 
 pub fn get_encoded_call(function: StrategyFunction, cellar_id: String) -> Result<Vec<u8>, Error> {
+    get_call(function, cellar_id).map(|call| call.encode())
```
EricBolten, line 28: Why do we need a map here? Isn't it always a singular call?

steward/src/commands/simulate.rs
```diff
@@ -0,0 +1,87 @@
...
+    fn run(&self) {
+        let config = APP.config();
+
+        info!("Starting application");
+        abscissa_tokio::run(&APP, async {
+            validate_tenderly_config(&config);
+
+            // Reflection required for certain clients to function... such as grpcurl
+            let contents = server::DESCRIPTOR.to_vec();
+            let proto_descriptor_service = tonic_reflection::server::Builder::configure()
```
EricBolten, line 36: Not a huge issue, but it's a bit weird in retrospect to have a singular descriptor file that, having added simulation protos, shows a superset of messages of which some will always be unavailable in either mode.

steward/src/commands/simulate.rs
```diff
@@ -0,0 +1,87 @@
...
+            std::process::exit(1)
+        });
+    }
+}
+
+impl config::Override<StewardConfig> for SimulateCmd {
+    // Process the given command line options, overriding settings from
+    // a configuration file using explicit flags taken from command-line
+    // arguments.
+    fn override_config(&self, config: StewardConfig) -> Result<StewardConfig, FrameworkError> {
```
EricBolten: Does this actually do anything?

steward/src/server.rs
```diff
@@ -35,3 +37,30 @@ pub async fn load_server_config(
         address,
     })
 }
+
+pub async fn load_tls_config(
+    server_cert_path: &str,
+    server_key_path: &str,
+    client_ca_path: Option<String>,
+) -> Result<ServerTlsConfig, Error> {
+    let cert = tokio::fs::read(server_cert_path).await?;
```
EricBolten, line 39: This looks like a duplicate of the beginning of `load_server_config`. If it's meant to be usable elsewhere by breaking the functionality out, should we just call `load_tls_config` in `load_server_config`?

steward/src/simulate.rs
```diff
@@ -0,0 +1,86 @@
...
+
+        Ok(Response::new(SimulateResponse {
+            encoded_call,
+            response_body: response,
+        }))
+    }
+}
+
+pub fn get_string_encoded_call(request: ScheduleRequest) -> Result<String, Error> {
+    if request.call_data.is_none() {
```
EricBolten: I imagine all of the functions here before the string encoding are exactly the same as the other place we handle `ScheduleRequest`, can we split this out into one function called in multiple places?

steward/src/simulate.rs
```diff
@@ -0,0 +1,86 @@
...
+        }
+        CellarV2(call) => {
+            if call.function.is_none() {
+                return Err(ErrorKind::Http.context("empty function data").into());
+            }
+
+            cellar_v2::get_encoded_call(call.function.unwrap(), request.cellar_id)
+        }
+    }
+    .map(|b| bytes_to_hex_str(&b))
```
EricBolten: Similar question here, isn't this a single value? Why does it need a map?

steward/src/tenderly.rs
EricBolten: Bump.

steward/src/tenderly.rs
EricBolten: That depends on how the Tenderly API works. Does it return an HTTP error type, or a 200 with a body indicating there's an error? What happens if we experience an actual error at the HTTP level?

EricBolten (commented):
Some feedback.

steward/src/cellars/aave_v2_stablecoin.rs
cbrit: Yes, this returns `Vec<u8>` (not `Vec<Bytes>`) which is a single call

steward/src/commands/simulate.rs
cbrit: mmm no I think it was part of the command gen boilerplate

steward/src/server.rs
cbrit: heh, turns out `load_tls_config` isn't called by anything, so yeah that works

steward/src/simulate.rs
cbrit: how about 

```rust
pub fn get_string_encoded_call(request: ScheduleRequest) -> Result<String, Error> {
    get_encoded_call(request).map(|b| bytes_to_hex_str(&b))
}
```

steward/src/simulate.rs
cbrit: map just changes one type to another, it's not making a hashmap. This is turning the `Vec<u8>` from `get_encoded_call` into a `String`

steward/src/tenderly.rs
cbrit: reqwest's error type has no explicit way to get a code, and logging a DNS error as a sample I get
```
reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("poopoo.poop")), port: None, path: "/account/cbrit/project/Project/simulate", query: None, fragment: None }, source: hyper::Error(Connect, ConnectError("dns error", Custom { kind: Uncategorized, error: "failed to lookup address information: nodename nor servname provided, or not known" })) }
```
Simulation error returns 200 with the failed transaction/log as the body.

steward/src/cellars/aave_v2_stablecoin.rs
EricBolten: Right, but `map` is applied to elements of an iterator. Unless `get_call` is returning an array of calls to encode, isn't this going to be calling the `map` function on every element of the incoming `vec`? I'm confused as to why this would be working.

steward/src/config.rs
```diff
@@ -267,6 +268,7 @@ impl Default for SimulateSection {
             tenderly_access_key: "".to_string(),
             tenderly_project_name: "".to_string(),
             tenderly_username: "".to_string(),
+            gravity_address: "0x69592e6f9d21989a043646fe8225da2600e5a0f7".to_string(),
```
EricBolten: If we're going to include a hardcoded default here, we probably should have left `GRAVITY_ADDRESS` defined and used it here rather than a magic string.

steward/src/simulate.rs
```diff
@@ -40,47 +31,27 @@ impl steward::simulate_contract_call_server::SimulateContractCall for SimulateHa
...
-        CellarV2(call) => {
-            if call.function.is_none() {
-                return Err(ErrorKind::Http.context("empty function data").into());
-            }
-
-            cellar_v2::get_encoded_call(call.function.unwrap(), request.cellar_id)
-        }
-    }
-    .map(|b| bytes_to_hex_str(&b))
+    get_encoded_call(request).map(|b| bytes_to_hex_str(&b))
```
EricBolten, line 56: `map` is applied to elements of an iterator. Unless `get_encoded_call` is returning an array of calls to encode, isn't this going to be calling the `map` function on every element of the incoming` vec`? I'm confused as to why this would be working.

steward/src/cellars/aave_v2_stablecoin.rs
EricBolten: What happens when `map` is called on a `Result` type here without checking to see if it's `Error`?

steward/src/cellars/cellar_v1.rs
```diff
@@ -26,6 +25,10 @@ const CELLAR_NAME: &str = "CellarV1";
 const LOG_PREFIX: &str = CELLAR_NAME;
 
 pub fn get_encoded_call(function: StrategyFunction, cellar_id: String) -> Result<Vec<u8>, Error> {
+    get_call(function, cellar_id).map(|call| call.encode())
```
EricBolten, line 28: Same question here re: `Error` results.

steward/src/cellars/cellar_v2.rs
```diff
@@ -47,6 +47,10 @@ pub fn normalize_address(address: String) -> String {
 }
 
 pub fn get_encoded_call(function: StrategyFunction, cellar_id: String) -> Result<Vec<u8>, Error> {
+    get_call(function, cellar_id).map(|call| call.encode())
+}
+
+pub fn get_call(function: StrategyFunction, cellar_id: String) -> Result<CellarV2Calls, Error> {
```
EricBolten, line 53: And here.

steward/src/commands/simulate.rs
EricBolten: Any thoughts on this? Should we have separate descriptor files for the different modes, as they will have different APIs?

steward/src/commands/simulate.rs
```diff
@@ -0,0 +1,76 @@
...
+impl Runnable for SimulateCmd {
+    /// Simulate the application.
+    fn run(&self) {
+        let config = APP.config();
+
+        info!("Starting application");
+        abscissa_tokio::run(&APP, async {
+            validate_tenderly_config(&config);
+
+            // Reflection required for certain clients to function... such as grpcurl
```
EricBolten: This isn't required for `grpcurl` to function, but it is required to use the `list` or `describe` commands with `grpcurl`.

steward/src/simulate.rs
EricBolten: Bumping this to note that here also, `map` is meant to apply to each element of an iterator, for a `Vec<u8>` I imagine that's each byte in the `vec`.

EricBolten (commented):
Some further comments and questions.

steward/src/simulate.rs
cbrit: It's opaque but this is a case of the `Result<T, E>` type's `map` function which simply changes the `T` to another type `U`:

https://doc.rust-lang.org/std/result/enum.Result.html#method.map

The T in this case is the vec, the whole of which gets passed into `bytes_to_hex_str`, not each item individually, in this case.

steward/src/cellars/cellar_v1.rs
cbrit: The result remains unwrapped

steward/src/commands/simulate.rs
cbrit: i.e. split it into two proto packages?

EDIT: it doesn't appear that you can create partial `FileDescriptorSet`s of a package.

steward/src/simulate.rs
cbrit: Noted elsewhere, map on `Result` behaves differently

EricBolten (approved):
LGTM

PR #187, opened 2023-02-09
Status endpoint

### LINKED ISSUE #180: Steward status endpoint

cbrit:
Closes #180 

Currently only returns the binary version. What other health check info should we include?


docs/api/adaptors/uniswap/uniswap_v3.md
```diff
@@ -7,6 +7,7 @@
     - [UniswapV3Adaptor](#steward-v3-UniswapV3Adaptor)
     - [UniswapV3Adaptor.AddToPosition](#steward-v3-UniswapV3Adaptor-AddToPosition)
     - [UniswapV3Adaptor.ClosePosition](#steward-v3-UniswapV3Adaptor-ClosePosition)
+    - [UniswapV3Adaptor.CollectFees](#steward-v3-UniswapV3Adaptor-CollectFees)
```
EricBolten, line 10: What's going on here? Did we forget to regenerate the docs when porting forward from 2.x?

proto/steward.proto
```diff
@@ -41,3 +42,12 @@ message ScheduleResponse {
...
     string id = 1;
 }
+
+/*
+ * Represents a request for Steward's current status
+ */
+message StatusRequest {}
+
+message StatusResponse {
+    string version = 1;
```
EricBolten, line 52: Are there other health indicators we may want to include? It doesn't necessarily mean we have to do it now, but some things to consider: last successful update times on some of the background threads, possibly some kind of error counts or flags, process start time/uptime, etc.

steward/src/commands/start.rs
```diff
@@ -30,8 +30,8 @@ impl Runnable for StartCmd {
         info!("Starting application");
         abscissa_tokio::run(&APP, async {
             // currently allows the threads to detach since we aren't capturing the JoinHandle
-            start_approved_cellar_cache_thread().await;
-            start_scheduled_cork_proposal_polling_thread().await;
+            // start_approved_cellar_cache_thread().await;
+            // start_scheduled_cork_proposal_polling_thread().await;
```
EricBolten: I assume commenting out the two above lines was for testing or accidental? Seems like we'd want to keep them.

steward/src/cork.rs
```diff
@@ -79,6 +87,12 @@ impl steward::contract_call_server::ContractCall for CorkHandler {
             id: id_hash(height, &cellar_id, encoded_call),
         }))
     }
+
+    async fn status(&self, _: Request<StatusRequest>) -> Result<Response<StatusResponse>, Status> {
+        Ok(Response::new(StatusResponse {
+            version: STEWARD_VERSION.to_string(),
```
EricBolten, line 93: Can we store the static value as a `String` so we don't have to convert it every time?

EricBolten (commented):
A few small questions.

steward/src/cork.rs
```diff
@@ -31,6 +35,10 @@ pub mod proposals;
 const MESSAGE_TIMEOUT: Duration = Duration::from_secs(10);
 const CHAIN_PREFIX: &str = "somm";
 
+lazy_static! {
+    static ref STEWARD_VERSION: &'static str = env!("CARGO_PKG_VERSION");
+};
```
EricBolten: I believe all the tests are failing because this line doesn't compile. We should fix and verify that the status endpoint works in practice.

docs/api/adaptors/uniswap/uniswap_v3.md
cbrit: yes

proto/steward.proto
cbrit: I was thinking along similar lines. In the future plz.

steward/src/cork.rs
cbrit: With the way lazy_static works we'd still need the `.to_string()` call

cbrit: Oh you mean store it separately from the static, I can do that

cbrit: We would still have to clone it each time to avoid moving out of the stored location, which is probably a similar cost right?

steward/src/cork.rs
cbrit: Sorry, was moving too fast. Should have caught this.

EricBolten (approved):
LGTM

PR #189, opened 2023-02-21
Build improvements and paying down related tech debt

### LINKED ISSUE #184: Steward CI performance improvements

### LINKED ISSUE #188: Use build.rs buildscript to generate ABI and protos
https://doc.rust-lang.org/cargo/reference/build-scripts.html

cbrit:
Lots going on here, I recommend reviewing one commit at a time. This PR will
- Fix build caching in CI to speed up builds and combine format, linting, building, and testing into one job. Second part is debatable. Felt weird having a bunch of repeated caching code for the separate jobs.
- Remove the workspace and various ABI/proto build/bindings crates in favor of using `build.rs` to build ABI and proto bindings. The generation runs before each build and drops the bindings directly into the steward crate. This could be improved by doing some sort of checksum to avoid unnecessary re-generation which can slow down builds by a few seconds (probably more in CI). 
- Move the contents of `steward/` into the root of the repository as a consequence of the above
- Reduce docker image size *a lot* by copying the final binary to a vanilla `debian` image (the same base image/version as the `rust` image used). Worth noting that a consequence of the change to use `build.rs` for proto bindings is that the version of `tonic-build` we are using requires `rustfmt` to be installed by default. 
- Speed up API doc build which now outputs a single markdown file
- ~~Build Go protos and API docs in the build.rs script~~

Other tidying:
- Move scripts into a `scripts` folder

Closes #184, closes #188

Note: The API docs now build to a single markdown file. If we want adaptors to be neatly separated we'll need put them in their own sub-package.


cbrit:
Moved to draft as there are some things that need tweaking

EricBolten:
Will review pending merge conflict resolution.

.github/workflows/automated_release.yml
```diff
@@ -24,18 +24,17 @@ jobs:
...
+        id: rust-cache
         with:
           path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            steward/target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('steward/Cargo.lock') }}
+            ~/.cargo/registry
```
EricBolten: Are there elements of the registry or git directories we might not want to cache, given the specific directory callouts in the previous version?

.github/workflows/integration_tests.yml
```diff
@@ -81,27 +82,24 @@ jobs:
...
-        with:
-          registry: ${{ env.REGISTRY }}
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-      - name: labels
-        id: meta
-        uses: docker/metadata-action@v3
-        with:
-          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+      # To avoid an image push, which requires a PR author to have write permissions
```
EricBolten: Appreciate the context but don't think we need the comment in the workflow.

.github/workflows/integration_tests.yml
```diff
@@ -167,10 +160,14 @@ jobs:
...
       - name: rename orchestrator image
         run: docker tag ${{ env.ORCHESTRATOR_IMAGE_NAME }} orchestrator:prebuilt
-      - name: pull steward image
-        run: docker pull ${{ steps.meta-steward.outputs.tags }}
-      - name: rename steward image
-        run: docker tag ${{ steps.meta-steward.outputs.tags }} steward:prebuilt
+      - name: Get steward artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ env.IMAGE_NAME }}
```
EricBolten, line 182: I'm confused, if `name` here is the file in `path`, wouldn't this need to be `env.STEWARD_ARTIFACT`?

.github/workflows/rust.yml
```diff
@@ -10,70 +9,31 @@ env:
...
-      - name: Run steward acceptance tests
-        run: cargo test --all --verbose
-
-  rust-build:
-    runs-on: ubuntu-20.04
-    steps:
-      - name: Checkout branch
-        uses: actions/checkout@v2
+          profile: minimal
+          toolchain: stable
```
EricBolten: I recall seeing the toolchain value marked as deprecated in messages, is this not the case? Also, I think we would benefit from pinning the rust toolchain to the expected version across jobs rather than using the latest stable runtime.

.github/workflows/rust.yml
```diff
@@ -10,70 +9,31 @@ env:
...
-        uses: actions/checkout@v2
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - name: Add rustfmt
```
EricBolten, line 57: I think these were separated out because it makes it clearer in Github Actions which aspect failed, rather than having to look at the logs. We might want to keep it that way.

Dockerfile
```diff
@@ -1,10 +1,11 @@
 # Reference: https://www.lpalmieri.com/posts/fast-rust-docker-builds/
 
-FROM rust:1.63 as cargo-chef-rust
+FROM rust:1.63-bullseye as cargo-chef-rust
```
EricBolten, line 3: What's this change doing?

Dockerfile
```diff
@@ -13,20 +14,19 @@ COPY . .
...
+WORKDIR /app
 COPY . .
 # Copy over the cached dependencies
 COPY --from=cacher /app/target target
 COPY --from=cacher /usr/local/cargo /usr/local/cargo
 RUN cargo build --release --bin steward
 
-FROM cargo-chef-rust as runtime
-WORKDIR app
+FROM debian:bullseye as runtime
```
EricBolten, line 29: Why are we pulling from `cargo-chef-rust` for every other command but not here?

Makefile
```diff
@@ -3,8 +3,9 @@
 VALIDATOR_IMAGE := "ghcr.io/peggyjv/sommelier-sommelier:main"
 ORCHESTRATOR_IMAGE := "ghcr.io/peggyjv/gravity-bridge-orchestrator:latest"
 
-build_protos:
-	./build_protos.sh
+protos:
```
EricBolten: The old `build_protos` script ran the rust proto build script as well, wouldn't we want the `protos` Makefile command to do that?

Makefile
```diff
@@ -15,27 +16,27 @@ e2e_build_images: e2e_clean_slate
...
 e2e_cork_test: e2e_aave_v2_stablecoin_test e2e_cellar_v1_test e2e_cellar_v2_test e2e_proposal_test
 
-# Because of the way `make` works, using the e2e_clean_slate target twice as prerequisites
-# for the individual tests doesn't work when `e2e_cork_test` runs the test targets in series,
+# Because of the way `make` works, using the e2e_clean_slate as as a prerequisite for
+# the individual tests doesn't work when `e2e_cork_test` runs the test targets in series,
 # so we explicitly call the cleanup script in each test target.
-e2e_aave_v2_stablecoin_test:
-	@./clean_slate.sh
+e2e_aave_v2_stablecoin_test: protos
```
EricBolten: Why are we building protos every time we run a test? This should only have to happen at dev time.

build.rs
```diff
@@ -16,25 +12,63 @@ use walkdir::WalkDir;
...
-const OUT_PATH: &str = "../steward_proto_rust/src/prost/";
-
-// All paths must end with a / and either be absolute or include a ./ to reference the current
-// working directory.
+const OUT_PATH: &str = "src/gen/proto/";
 
 fn main() {
-    let out_dir = Path::new(&OUT_PATH);
-    let tmp_dir = Path::new(&TMP_PATH);
+    generate_contract_abis();
```
EricBolten, line 18: Perhaps the Makefile should execute this `main` function, which would fix the lack of rust generation in the `protos` command and consolidate the script usage to one place.

EricBolten (commented):
Some comments and questions.

.github/workflows/integration_tests.yml
cbrit: It turns out that name is not the file name

Dockerfile
cbrit: To minimize the final image size. `cargo-chef-rust` pulls in everything in the `rust` image which is only necessary for the build phase.

Dockerfile
cbrit: I think it was to make sure I was using the same OS in the final runtime image and the build image.

Makefile
cbrit: It's done by the rust `build.rs` automatically on build now

Makefile
cbrit: Since I moved the rust proto build into the `build.rs` file which executes every build, I figured it the same spirit I would make this change. It may be unnecessary but seemed like a good idea at the time.

.github/workflows/rust.yml
mvid: It also allows them to run in parallel if they aren't depending on each other

.github/workflows/rust.yml
```diff
@@ -1,39 +1,75 @@
 name: Rust tests
+
 on:
   push:
     branches:
-      - main
```
EricBolten: I don't think we want to remove these four lines.

Makefile
```diff
@@ -4,8 +4,7 @@ VALIDATOR_IMAGE := "ghcr.io/peggyjv/sommelier-sommelier:main"
 ORCHESTRATOR_IMAGE := "ghcr.io/peggyjv/gravity-bridge-orchestrator:latest"
 
 protos:
-	scripts/build_go_protos.sh
-	scripts/build_api_docs.sh
+	@cargo check --locked
```
EricBolten: This might benefit from a comment explaining why `cargo check` builds the protos.

build.rs
```diff
@@ -27,26 +27,11 @@ fn generate_contract_abis() {
         ("AaveV2StablecoinCellar", "aave_v2_stablecoin"),
         ("CellarV1", "cellar_v1"),
         ("CellarV2", "cellar_v2"),
-        ("CellarV2_2", "cellar_v2_2"),
```
EricBolten: Why are we deleting all of these that was just added earlier? Also why delete `abi/CellarV2_2.json`?

.github/workflows/rust.yml
```diff
@@ -20,12 +16,12 @@ jobs:
...
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+            ~/.cargo/bin
+            ~/.cargo/registry/index
+            ~/.cargo/registry/cache
+            ~/.cargo/git/db
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
EricBolten: Is there more than one `Cargo.lock` now that the workspace has been removed?

Cargo.toml
```diff
@@ -1,30 +1,77 @@
...
-	"steward_proto_rust_build",
-]
-members = [
-	"steward",
-	"steward_abi",
-	"steward_abi_build",
-	"steward_proto_rust",
-	"steward_proto_rust_build",
-]
+edition = "2018"
```
EricBolten: Why are we downgrading rust edition? Is this related to importing the orchestrator dependencies?

Dockerfile
EricBolten: Based on the end of this file it seems like `bullseye` might target a Debian image, but we do our operations in Github actions using Ubuntu 20.04. Would there be benefit in being consistent here?

Makefile
EricBolten: I suppose, but this means CI will be building protos, and if the build introduces file changes that weren't committed, CI results won't match the checked-in code.

src/utils.rs
```diff
@@ -84,19 +76,6 @@ pub async fn get_chain(eth_client: Provider<Http>) -> Result<Chain, Error> {
     })
 }
 
-pub async fn get_delegates_keys_by_orchestrator(
```
EricBolten, line 87: Was nothing calling this then?

.github/workflows/rust.yml
```diff
@@ -1,39 +1,75 @@
...
+
+  fmt:
+    name: Rustfmt
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+        with:
+          profile: minimal
+          toolchain: stable
```
EricBolten: This action uses `with: profile: toolchain:`, as does `clippy`, but the others do not. We should either explicitly include them all using the toolchain of the rust version we expect rather than `stable`, or we should remove them all if it turns out they are unnecessary (they might have been deperecated?).

EricBolten (commented):
Some follow-up comments.

EricBolten:
It also appears the `steward-build` action was broken in this changeset, building the go protos is panicking.

build.rs
cbrit: This PR doesn't have handlers for new cellars yet, I think I originally added this by mistake, though I could just be confused because of the split main lines state right now

src/utils.rs
cbrit: I think it was leftover from an old command

.github/workflows/automated_release.yml
```diff
@@ -24,18 +24,19 @@ jobs:
...
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            steward/target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('steward/Cargo.lock') }}
+            ~/.cargo/bin
+            ~/.cargo/registry/index
+            ~/.cargo/registry/cache
+            ~/.cargo/git/db
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
EricBolten: Since this has moved out of a workspace, is there more than one `Cargo.lock` to care about?

.github/workflows/integration_tests.yml
```diff
@@ -81,27 +82,21 @@ jobs:
...
+          push: false
+          tags: steward:prebuilt
           builder: ${{ steps.buildx.outputs.name }}
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,mode=max,dest=/tmp/.buildx-cache-new
+          outputs: type=docker,dest=/tmp/${{ env.STEWARD_ARTIFACT }}
+      - name: Make Steward artifact available to other jobs
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ env.IMAGE_NAME }}
```
EricBolten, line 113: Would this cause concurrent CI jobs to overwrite the same artifact and possibly create conflicts? I would think we would need to tag the name and path with a version or PR number, in a similar way that we had to for docker images.

build.rs
```diff
@@ -16,27 +12,63 @@ use walkdir::WalkDir;
...
-    println!(
-        "Compiling .proto files to Rust into '{}'...",
-        out_dir.display()
-    );
+fn generate_contract_abis() {
+    // (JSON/Contract name, output file name)
+    let contracts = vec![
+        ("AaveV2StablecoinCellar", "aave_v2_stablecoin"),
+        ("CellarV1", "cellar_v1"),
+        ("CellarV2", "cellar_v2"),
```
EricBolten, line 27: Just double checking, at this point we don't have Cellar 2.2 yet, that's coming when we merge the latest release into main, yes? Will also be lots more adaptors to add.

proto/steward/v3/steward.proto
```diff
@@ -16,7 +16,7 @@ import "cellar_v2.proto";
 /*
  * Service for handling Cellar contract calls
  */
-service ContractCall {
+service ContractCallService {
```
EricBolten: Renaming the service is a GRPC API-breaking change. We will need to either leave it the same or bump the major version of the API.

EricBolten (commented):
A few more suggestions and questions.

build.rs
cbrit: Yes, those will be included in the catch-up work

.github/workflows/integration_tests.yml
```diff
@@ -95,7 +96,7 @@ jobs:
       - name: Make Steward artifact available to other jobs
         uses: actions/upload-artifact@v3
         with:
-          name: ${{ env.IMAGE_NAME }}
+          name: ${{ env.IMAGE_NAME }}-pr-${{ env.PR_NUMBER }}
```
EricBolten: I feel like we talked about this before, but do we need to include the PR number in the path as well to ensure it's unique?

build.rs
```diff
@@ -68,7 +68,7 @@ fn generate_rust_protos() {
     let root = env!("CARGO_MANIFEST_DIR");
     let root: PathBuf = root.parse().unwrap();
     let mut steward_proto_dir = root;
-    steward_proto_dir.push("proto/steward/v3");
+    steward_proto_dir.push("proto/steward/v4");
```
EricBolten, line 71: We'll need to update the inclusion of the v4 generated rust file, it's breaking the rust build.

EricBolten (commented):
Some final suggestions.

.github/workflows/integration_tests.yml
cbrit: my bad

EricBolten (approved):
LGTM

PR #198, opened 2023-03-29
Include relevant changes from the 2.x line

cbrit:
Includes changes that seem relevant to 3.0.0 from the PR to 2.x-main #192 

Things I left out:

- sommelier/v4/ -> /v5/ go package change
- `SetupAdaptor` strategist function and related APPROVED_ADAPTOR const


proto/adaptors/uniswap/uniswap_v3.proto
```diff
@@ -29,6 +29,12 @@ message UniswapV3Adaptor {
         TakeFromPosition take_from_position = 7;
         // Represents function `collectFees(uint256 positionId, uint128 amount0, uint128 amount1)`
         CollectFees collect_fees = 8;
+        // Represents function `purgeAllZeroLiquidityPositions(ERC20 token0, ERC20 token1)`
+        PurgeAllZeroLiquidityPositions purge_all_zero_liquidity_positions = 9;
```
EricBolten, line 33: I think the 2.x line failed to do this, but we should probably re-run the doc generation.

EricBolten (commented):
Aside from the comment, I think we're mistakenly including `steward_abi/src/cellar.rs` here. The 2.x line also contained updates moving our integration test version of somm from v4 to v5, although I guess we might not be able to do that as main is pinned to a commit on somm main with the cork changes in it?

cbrit:
> Aside from the comment, I think we're mistakenly including `steward_abi/src/cellar.rs` here. The 2.x line also contained updates moving our integration test version of somm from v4 to v5, although I guess we might not be able to do that as main is pinned to a commit on somm main with the cork changes in it?

On integration tests: yeah, I couldn't upgrade because there are steward tests for cork v2/auction/proposal scheduled corks.

Fixed the other two issues!

EricBolten (approved):
LGTM

PR #200, opened 2023-04-11
Cellar v2.2 steward support, upgraded adaptor support

cbrit:


proto/cellar_v2.proto
```diff
@@ -57,25 +64,138 @@ message CellarV2 {
...
+        // Index at which to remove the position
+        uint32 index = 1;
+        // Whether to remove position from the debt array, or the credit array.
+        bool in_debt_array = 2;
+    }
+
+    /*
+    * Set the holding position used of the cellar.
+    *
+    * Represents function `setHoldingIndex(uint8 index)`
```
EricBolten: This doesn't match the proto described below, can we verify the expected correct one?

proto/cellar_v2.proto
```diff
@@ -57,25 +64,138 @@ message CellarV2 {
...
+    * Represents function `setStrategistPayoutAddress(address payout)`
+    */
+    message SetStrategistPayoutAddress {
+        string payout = 1;
+    }
+
+    /*
+    * Swap the positions at two given indeces.
+    *
+    * Represents function `swapPositions(uint32 index1, uint32 index2)`
```
EricBolten: Also verify this, the comment does not describe a bool field which is included below.

proto/cellar_v2.proto
```diff
@@ -57,25 +64,138 @@ message CellarV2 {
...
+        // Index of the second position
+        uint32 index_2 = 2;
+        // Whether to switch positions in the debt array, or the credit array.
+        bool in_debt_array = 3;
+    }
+
+    /*
+    * Allows share lock period to be updated.
+    *
+    * Represents function `setShareLockPeriod()`
```
EricBolten: Another mismatch here -- these might have been ported from the existing CellarV2 calls, just trying to be thorough.

proto/cellar_v2.proto
```diff
@@ -150,11 +270,57 @@ message CellarV2 {
...
     /*
-    * Allows owner to add new adaptors for the cellar to use. They must have already been approved by the cellar's registry.
+    * Shutdown the cellar. Used in an emergency or if the cellar has been deprecated.
     *
-    * Represents function `setupAdaptor(address _adaptor)`
+    * Represents function `initiateShutdown()`
     */
-    message SetupAdaptor {
-        string adaptor = 1;
+    message InitiateShutdown {}
```
EricBolten, line 311: I don't remember where we landed on this -- are we allowing the strategist to call this unilaterally?

steward/src/cellars.rs
```diff
@@ -5,8 +5,37 @@ use std::result::Result;
...
 pub(crate) mod cellar_v1;
 pub(crate) mod cellar_v2;
+pub(crate) mod cellar_v2_2;
+
+// adaptors and positions associated with the deprecated UniV3 adaptor are blocked
+// addresses treated as lowercase without 0x prefix to ensure valid comparisons with arbitrary input
+pub const BLOCKED_ADAPTORS: [&str; 1] = ["7c4262f83e6775d6ff6fe8d9ab268611ed9d13ee"];
+pub const BLOCKED_POSITIONS: [u32; 9] = [4, 5, 6, 7, 8, 9, 10, 11, 12];
+pub const ALLOWED_SETUP_ADAPTORS: [&str; 9] = [
+    // UniswapV3Adaptr V2
```
EricBolten: I'm a bit confused by this -- I don't see a UniswapV3AdaptorV2 in the AdaptorCall list.

Nit: Adaptr -> Adaptor

steward/src/cellars.rs
```diff
@@ -5,8 +5,37 @@ use std::result::Result;
...
 
 pub(crate) mod aave_v2_stablecoin;
+pub(crate) mod adaptors;
 pub(crate) mod cellar_v1;
 pub(crate) mod cellar_v2;
+pub(crate) mod cellar_v2_2;
+
+// adaptors and positions associated with the deprecated UniV3 adaptor are blocked
+// addresses treated as lowercase without 0x prefix to ensure valid comparisons with arbitrary input
+pub const BLOCKED_ADAPTORS: [&str; 1] = ["7c4262f83e6775d6ff6fe8d9ab268611ed9d13ee"];
```
EricBolten: This contains the old Uniswap adaptor address, but we need to add all the replaced adaptors reflected by the new blocked positions.

steward/src/cellars.rs
```diff
@@ -5,8 +5,37 @@ use std::result::Result;
...
+pub const BLOCKED_POSITIONS: [u32; 9] = [4, 5, 6, 7, 8, 9, 10, 11, 12];
+pub const ALLOWED_SETUP_ADAPTORS: [&str; 9] = [
+    // UniswapV3Adaptr V2
+    "dbd750f72a00d01f209ffc6c75e80301efc789c1",
+    // VestingSimpleAdaptor V2
+    "508E6aE090eA92Cb90571e4269B799257CD78CA1",
```
EricBolten: Because of the way the address normalization works, these hardcoded values have to already be all lowercase.

integration_tests/cork_test.go
```diff
@@ -633,6 +633,121 @@ func (s *IntegrationTestSuite) TestCellarV2() {
...
+		cellarId := v2_2Cellar.Hex()
+
+		// Contains two adaptor calls, the first of which has two function calls inside of it, for a total of three function calls.
+		request := &steward_proto.SubmitRequest{
+			CellarId: cellarId,
+			CallData: &steward_proto.SubmitRequest_CellarV2_2{
+				CellarV2_2: &steward_proto.CellarV2_2{
+					CallType: &steward_proto.CellarV2_2_Multicall_{
+						Multicall: &steward_proto.CellarV2_2_Multicall{
+							FunctionCalls: []*steward_proto.CellarV2_2_FunctionCall{
```
EricBolten, line 648: Do we want to include more than one `FunctionCall` when testing `Multicall` here?

integration_tests/cork_test.go
```diff
@@ -633,6 +633,121 @@ func (s *IntegrationTestSuite) TestCellarV2() {
...
+			s.T().Logf("got %v logs: %v", len(logs), logs)
+			if len(logs) > 0 {
+				for _, log := range logs {
+					if len(log.Data) > 0 {
+						var event AdaptorSwapWithUniV3
+						err := adaptor_abi.UnpackIntoInterface(&event, "SwapWithUniV3", log.Data)
+						s.Require().NoError(err, "failed to unpack SwapWithUniV3 event from log data")
+						s.Require().Equal(big.NewInt(3), event.AmountOutMin)
+
+						s.T().Log("Saw BorrowFromAave event!")
```
EricBolten: Need to update this log line.

integration_tests/cork_test.go
```diff
@@ -633,6 +633,121 @@ func (s *IntegrationTestSuite) TestCellarV2() {
...
+			}
+
+			s.T().Logf("got %v logs: %v", len(logs), logs)
+			if len(logs) > 0 {
+				for _, log := range logs {
+					if len(log.Data) > 0 {
+						var event AdaptorSwapWithUniV3
+						err := adaptor_abi.UnpackIntoInterface(&event, "SwapWithUniV3", log.Data)
+						s.Require().NoError(err, "failed to unpack SwapWithUniV3 event from log data")
+						s.Require().Equal(big.NewInt(3), event.AmountOutMin)
```
EricBolten, line 761: Should we be checking the other fields as well?

integration_tests/cork_test.go
```diff
@@ -633,6 +633,121 @@ func (s *IntegrationTestSuite) TestCellarV2() {
 	})
 }
 
+func (s *IntegrationTestSuite) TestCellarV2_2() {
+	s.Run("Submit callOnAdaptor to MockCellar", func() {
+		s.checkCellarExists(vaultCellar)
+		s.waitForVotePeriod()
+
+		// val := s.chain.validators[0]
```
EricBolten: Let's drop this if it's unused.

integration_tests/cork_test.go
```diff
@@ -633,6 +633,121 @@ func (s *IntegrationTestSuite) TestCellarV2() {
...
+						},
+					},
+				},
+			},
+		}
+
+		s.waitForVotePeriod()
+		s.executeStewardCalls(request)
+		s.waitForVotePeriod()
+		// s.queryLogicCallTransactionByAddress(clientCtx, v2_2Cellar.Hex())
```
EricBolten: Our other tests are running this, but we've commented this out here. Is it broken?

proto/adaptors/0x/zero_x.proto
```diff
@@ -0,0 +1,38 @@
...
+    /*
+    * Allows strategists to make ERC20 swaps using 0x.
+    *
+    * Represents function `swapWith0x(ERC20 tokenIn, ERC20 tokenOut, uint256 amount, bytes memory swapCallData)`
+    */
+    message SwapWith0x {
+        string token_in = 1;
+        string token_out = 2;
+        string amount = 3;
+        bytes swap_call_data = 4;
```
EricBolten, line 32: For 0x, OneInch, etc., should we be encoding the `swap_call_data` similar to how we handle e.g. Uniswap?

proto/adaptors/1inch/oneinch.proto
```diff
@@ -0,0 +1,32 @@
...
+syntax = "proto3";
+package steward.v3;
+
+option go_package = "/steward_proto";
+
+// Represents call data for the OneInch adaptor.
+message OneInchAdaptorV1 {
+    oneof function {
+        // Represents function `swapWithOneInch(ERC20 tokenIn, ERC20 tokenOut, uint256 amount, bytes swapCallData)`
+        SwapWithOneInch swap_with_one_inch = 1;
```
EricBolten: It seems like most adaptors contain `RevokeApproval`, should that be present here as well?

proto/adaptors/aave/aave_v3_a_token.proto
```diff
@@ -0,0 +1,81 @@
...
+
+import "adaptors/base.proto";
+
+// Represents call data for the Aave AToken adaptor, used to manage lending positions on Aave
+message AaveV3ATokenAdaptorV1 {
+    oneof function {
+        /***** BASE ADAPTOR FUNCTIONS *****/
+
+        // Represents function `revokeApproval(ERC20 asset, address spender)`
+        RevokeApproval revoke_approval = 3;
```
EricBolten: Since this is a new proto file we could start the index at 1.

proto/adaptors/aave/aave_v3_debt_token.proto
```diff
@@ -0,0 +1,135 @@
...
+    message FlashLoan {
+        // The addresses of the ERC20 tokens to borrow
+        repeated string loan_tokens = 1;
+        // The amounts to borrow
+        repeated string loan_amounts = 2;
+        // The params to pass to the flash loan callback.
+        repeated AdaptorCallForAaveV3Flashloan params = 3;
+    }
+
+    // NOTE: FlashLoan takes an array of AdaptorCall. cellar_v2.proto defines it, but also imports this file, therefore we can't import cellar_v2.proto in order to use the AdaptorCall message here. To avoid the recursive import, we duplicate the message definition.
```
EricBolten, line 93: Could we move AdaptorCall to a common file imported by both?

proto/adaptors/aave/aave_v3_debt_token.proto
```diff
@@ -0,0 +1,135 @@
...
+import "adaptors/base.proto";
+import "adaptors/sommelier/cellar_adaptor.proto";
+
+// Represents call data for the Aave Debt Token adaptor, used for borrowing and repaying debt on Aave.
+message AaveV3DebtTokenAdaptorV1 {
+    oneof function {
+        /***** BASE ADAPTOR FUNCTIONS *****/
+
+        // Represents function `revokeApproval(ERC20 asset, address spender)`
+        RevokeApproval revoke_approval = 3;
```
EricBolten: Same here re: indexing.

proto/adaptors/fees_and_reserves.proto
```diff
@@ -0,0 +1,188 @@
...
+    message PrepareFees {
+        string amount = 1;
+    }
+}
+
+message FeesAndReservesAdaptorV1Calls {
+    repeated FeesAndReservesAdaptorV1 calls = 1;
+}
+
+// /**
```
EricBolten: Looks like this was left here for reference, let's delete.

proto/steward.proto
```diff
@@ -29,9 +29,10 @@ message SubmitRequest {
     string cellar_id = 1;
     // The data from which the desired contract function will be encoded
     oneof call_data {
-        AaveV2Stablecoin aave_v2_stablecoin = 2;
-        CellarV1 cellar_v1 = 3;
-        CellarV2 cellar_v2 = 4;
+        AaveV2Stablecoin aave_v2_stablecoin = 3;
```
EricBolten: Why renumber these?

steward/src/cellars/adaptors/aave_v2.rs
```diff
@@ -0,0 +1,293 @@
...
+    params: steward_proto::steward::AaveDebtTokenAdaptorV2Calls,
+) -> Result<Vec<Bytes>, Error> {
+    let mut calls = Vec::new();
+    for c in params.calls {
+        let function = c
+            .function
+            .ok_or_else(|| sp_call_error("function cannot be empty".to_string()))?;
+
+        match function {
+            steward_proto::steward::aave_debt_token_adaptor_v2::Function::RevokeApproval(p) => {
```
EricBolten: Weird import inconsistency here with the rest of the file.

steward/src/cellars/adaptors/aave_v2.rs
```diff
@@ -0,0 +1,293 @@
...
+use ethers::{abi::AbiEncode, types::Bytes};
+use steward_abi::{
+    aave_a_token_adaptor::AaveATokenAdaptorCalls,
+    aave_a_token_adaptor_v2::AaveATokenAdaptorV2Calls,
+    aave_debt_token_adaptor::AaveDebtTokenAdaptorCalls,
+    aave_debt_token_adaptor_v2::AaveDebtTokenAdaptorV2Calls,
+};
+use steward_proto::steward::{
+    aave_a_token_adaptor_v1, aave_a_token_adaptor_v2, aave_debt_token_adaptor_v1,
+    AaveATokenAdaptorV1Calls, AaveDebtTokenAdaptorV1Calls,
```
EricBolten: Looking at this import structure, we seem to mix and match between "no version means v1" and explicitly specifying v1. It also seems like the import names in both `steward_abi` and `steward_proto` could clash because they're named exactly the same thing. Is there a way we could clean this up or make it more consistent?

steward/src/cellars/adaptors/aave_v3.rs
```diff
@@ -0,0 +1,216 @@
...
+                };
+                calls.push(
+                    AaveV3ATokenAdaptorCalls::AdjustIsolationModeAssetAsCollateral(call)
+                        .encode()
+                        .into(),
+                )
+            }
+            aave_v3a_token_adaptor_v1::Function::ChangeEmode(p) => {
+                let call = steward_abi::aave_v3_a_token_adaptor::ChangeEModeCall {
+                    category_id: p.category_id as u8,
```
EricBolten, line 80: I'm remembering we had to add `convert_exchange` with some enum because the numbers would be offset, any concern that could happen here?

steward/src/cellars/adaptors/aave_v3.rs
```diff
@@ -0,0 +1,216 @@
...
+                calls.push(AaveV3DebtTokenAdaptorCalls::FlashLoan(call).encode().into())
+            }
+        }
+    }
+
+    Ok(calls)
+}
+
+/// Encodes calls to the Adaptor contracts
+fn get_encoded_adaptor_call(
```
EricBolten: Nit: should this be plural?

steward/src/cellars/adaptors/uniswap_v3.rs
```diff
@@ -0,0 +1,219 @@
...
+                        .into(),
+                )
+            }
+        }
+    }
+
+    Ok(calls)
+}
+
+pub(crate) fn swap_with_uniswap_adaptor_v1_call(
```
EricBolten: This is a separate adaptor from the UniswapV3 adaptor, should we move it to its own file?

steward/src/cellars/adaptors/aave_v2.rs
```diff
@@ -0,0 +1,293 @@
...
+                )
+            }
+        }
+    }
+
+    Ok(calls)
+}
+
+pub(crate) fn aave_a_token_adaptor_v2_call(
+    params: steward_proto::steward::AaveATokenAdaptorV2Calls,
```
EricBolten, line 216: This is also present in a couple of places, where this is defined inline rather than as an import, makes the import list a tad confusing.

steward/src/cellars/cellar_v2.rs
```diff
@@ -185,503 +159,47 @@ fn get_encoded_adaptor_call(data: Vec<AdaptorCall>) -> Result<Vec<AbiAdaptorCall
...
-                            calls.push(
-                                UniswapV3AdaptorCalls::RemoveUnOwnedPositionFromTracker(call)
-                                    .encode()
-                                    .into(),
-                            )
-                        }
-                    }
-                }
+            UniswapV3Calls(params) => {
+                calls.extend(adaptors::uniswap_v3::uniswap_v3_adaptor_v1_call(params)?)
```
EricBolten: Extend is then pushing all of the bytes in the result into the `calls` vector of bytes? Was `push(...into())` doing this implicitly before? Just realizing how weird it is considering `calls` should only be one call ever.

steward/src/cellars/cellar_v2_2.rs
```diff
@@ -0,0 +1,244 @@
...
+                &SetRebalanceDeviationCall::function_name(),
+                &cellar_id,
+            );
+            let call = SetRebalanceDeviationCall {
+                new_deviation: string_to_u256(params.new_deviation)?,
+            };
+
+            Ok(CellarV2_2Calls::SetRebalanceDeviation(call).encode())
+        }
+        // Function::SetupAdaptor(params) => {
```
EricBolten: Did we land on needing to include this for 2.2?

steward/src/cellars/cellar_v2_2.rs
```diff
@@ -0,0 +1,244 @@
...
+            );
+            let call = InitiateShutdownCall {};
+
+            Ok(CellarV2_2Calls::InitiateShutdown(call).encode())
+        }
+    }
+}
+
+/// Encodes calls to the Adaptor contracts
+fn get_encoded_adaptor_call(data: Vec<AdaptorCall>) -> Result<Vec<AbiAdaptorCall>, Error> {
```
EricBolten: This function is exactly the same in both cellar v2 contexts, right? Can we split it out?

steward/src/cork.rs
```diff
@@ -131,6 +131,13 @@ fn get_encoded_call(request: SubmitRequest) -> Result<Vec<u8>, Error> {
 
             cellar_v2::get_encoded_call(call.function.unwrap(), request.cellar_id)
         }
+        CellarV22(call) => {
```
EricBolten: Is there any kind of separator we can use here? It makes theoretical semver clashes very possible.

EricBolten (commented):
Some comments and questions.

steward/src/cellars.rs
cbrit: This was one of the missing items i fixed in the latest push

steward/src/cellars.rs
cbrit: Added old vesting and compound addresses

steward/src/cellars.rs
cbrit: downvote

integration_tests/cork_test.go
cbrit: Yes, intended to, thanks.

integration_tests/cork_test.go
cbrit: Think it was a local bug because the other tests failed there too but CI is green

proto/adaptors/0x/zero_x.proto
cbrit: The respective platforms have a service that they will get the encoded bytes from and pass them in. There was a question about whether to make them encode it as raw bytes (since they get back a hex string) or have steward do it, I was going to have a look at doing it with steward and make these fields strings.

proto/cellar_v2.proto
```diff
@@ -46,6 +46,8 @@ message CellarV2 {
         SetShareLockPeriod set_share_lock_period = 8;
         // Represents function `setupAdaptor(address _adaptor)`
         SetupAdaptor setup_adaptor = 9;
+        // Represents function `setPlatformFee(uint64 platformFee)`
+        SetPlatformFee set_platform_fee = 10;
```
EricBolten, line 50: Why would we need to adjust the platform fee on v2? Wasn't this only for the new cellar? Was this meant to be in the 2.2 section?

proto/adaptors/aave/aave_v3_debt_token.proto
cbrit: No, because it contains messages defined in each adaptor's proto file which must be imported by the common file, including aave_v3_debt_token.proto, which needs the type from the common file.

integration_tests/cork_test.go
```diff
@@ -566,13 +558,12 @@ func (s *IntegrationTestSuite) TestCellarV2() {
...
-						s.Require().NoError(err, "failed to unpack ClaimCompAndSwap event from log data")
-						// 1 == UniswapV3
-						s.Require().Equal(uint8(1), event.Exchange)
+						var event AdaptorSwapWithUniV3
+						err := adaptor_abi.UnpackIntoInterface(&event, "SwapWithUniV3", log.Data)
+						s.Require().NoError(err, "failed to unpack SwapWithUniV3 event from log data")
+						s.Require().Equal(big.NewInt(3), event.AmountOutMin)
 
-						s.T().Log("Saw ClaimCompAndSwap event!")
+						s.T().Log("Saw BorrowFromAave event!")
```
EricBolten: Should fix this log line.

steward/src/cellars.rs
```diff
@@ -12,10 +12,17 @@ pub(crate) mod cellar_v2_2;
 
 // adaptors and positions associated with the deprecated UniV3 adaptor are blocked
 // addresses treated as lowercase without 0x prefix to ensure valid comparisons with arbitrary input
-pub const BLOCKED_ADAPTORS: [&str; 1] = ["7c4262f83e6775d6ff6fe8d9ab268611ed9d13ee"];
+pub const BLOCKED_ADAPTORS: [&str; 3] = [
+    // Original uniswap v3 adaptor
+    "7c4262f83e6775d6ff6fe8d9ab268611ed9d13ee",
+    // Original vesting simple adaptor
+    "1eAA1a100a460f46A2032f0402Bc01FE89FaAB60",
```
EricBolten: The new blocked adaptors also need to be lowercased.

steward/src/cellars/cellar_v2.rs
```diff
@@ -145,6 +145,28 @@ pub fn get_encoded_call(function: StrategyFunction, cellar_id: String) -> Result
...
             Ok(CellarV2Calls::SetupAdaptor(call).encode())
         }
+        SetPlatformFee(params) => {
+            log_cellar_call(
+                CELLAR_NAME,
+                &SetPlatformFeeCall::function_name(),
+                &cellar_id,
+            );
+
+            if params.platform_fee != 0 {
```
EricBolten: Where were we intending to set the platform fee to zero? Additionally, this would need to be tied to specific cellar addresses, we don't want to be able to set the platform fee to zero on any cellar.

steward/src/cellars/cellar_v2_2.rs
```diff
@@ -145,27 +149,19 @@ pub fn get_encoded_function(call: FunctionCall, cellar_id: String) -> Result<Vec
                 &SetRebalanceDeviationCall::function_name(),
                 &cellar_id,
             );
-            let call = SetRebalanceDeviationCall {
-                new_deviation: string_to_u256(params.new_deviation)?,
-            };
+
+            // bullet proof anti-rug
```
EricBolten: Funny but not useful.

steward/src/cellars/cellar_v2_2.rs
```diff
@@ -176,6 +172,18 @@ pub fn get_encoded_function(call: FunctionCall, cellar_id: String) -> Result<Vec
 
             Ok(CellarV2_2Calls::InitiateShutdown(call).encode())
         }
+        Function::SetStrategistPlatformCut(params) => {
+            log_cellar_call(
+                CELLAR_NAME,
+                &SetStrategistPlatformCutCall::function_name(),
+                &cellar_id,
+            );
+            let call = SetStrategistPlatformCutCall {
```
EricBolten, line 197: Strategist cut and platform fee both need specific values and specific cellar addresses.

EricBolten (commented):
Some follow-up comments.

steward/src/cellars/adaptors/aave_v3.rs
cbrit: In this case the actual param type in the Solidity is uint8, not enum, and we make the proto field an int as well, so no shifting is necessary.

steward/src/cellars/adaptors/aave_v2.rs
cbrit: I had decided that for adaptors whose V1 was now blocked, I'd just leave it unversioned, but you're right it's inconsistent. Fixed that.

As far as the name overlapping, I think it would be a good tech debt issue. There are a lot of places where I didn't use an in-line import path and you can't immediately tell which type it is.

cbrit: Update: I made all of the ABI file and type names consistent with the versioning, and then aliased the imports with the `Abi*` prefix.

steward/src/cellars/cellar_v2.rs
cbrit: Before we were looping inside of each enum variant, pushing each iteration. This does bring to my attention the fact that I named these methods as "call" instead of "calls". They return a vec so I'll rename them. That's also why `extend` is used.

steward/src/cellars/cellar_v2_2.rs
cbrit: This is one of those duplications that the build script PR will fix. 

The different cellar v2 definitions have their own AdaptorCall types, requiring two of these functions. The build PR makes the ABI definitions part of the steward crate, which would allow me to impl `From<T>` between the types and have a single function.

steward/src/cellars.rs
```diff
@@ -16,30 +16,30 @@ pub const BLOCKED_ADAPTORS: [&str; 3] = [
...
-    "1eAA1a100a460f46A2032f0402Bc01FE89FaAB60",
+    "1eaa1a100a460f46a2032f0402bc01fe89faab60",
     // Original compound C token adaptor
-    "26DbA82495f6189DDe7648Ae88bEAd46C402F078",
+    "26dba82495f6189dde7648ae88bead46c402f078",
 ];
 pub const BLOCKED_POSITIONS: [u32; 9] = [4, 5, 6, 7, 8, 9, 10, 11, 12];
 pub const ALLOWED_SETUP_ADAPTORS: [&str; 9] = [
-    // UniswapV3Adaptor V2
+    // Uniswapv3Adaptor v2
```
EricBolten: This address is unchanged -- was there a new UniswapV3 update to make it a V2 beyond the bugfix that went into the replacement last time? I believe the new one IIRC from this PR has some functions removed or changed, and if so, it's probably not the same address as what was originally here.

steward/src/cellars/adaptors/vesting_simple.rs
```diff
@@ -1,14 +1,14 @@
...
+use steward_proto::steward::vesting_simple_adaptor_v2;
 
 use crate::{
     error::Error,
     utils::{sp_call_error, sp_call_parse_address, string_to_u256},
 };
 
-pub(crate) fn vesting_simple_adaptor_v1_call(
-    params: steward_proto::steward::VestingSimpleAdaptorCalls,
+pub(crate) fn vesting_simple_adaptor_v2_calls(
```
EricBolten, line 10: We're renaming v1 to v2 here -- is the implication that nothing will ever need to call v1 again?

steward_abi_build/src/main.rs
```diff
@@ -9,22 +9,21 @@ fn main() {
         ("CellarV1", "cellar_v1"),
         ("CellarV2", "cellar_v2"),
         ("CellarV2_2", "cellar_v2_2"),
```
EricBolten, line 11: Is this affected by the 2dot2 change or is that just a proto thing?

EricBolten (commented):
A few more comments.

steward/src/cellars/adaptors/vesting_simple.rs
cbrit: yes, v1 had a vuln and is now blocked

steward/src/cellars.rs
cbrit: we checked the adaptor list and they said it was good, that's about as much as I know. Realizing now that we also kept the V1 proto in case someone wants to use it in the future but it's not in this list.

steward_abi_build/src/main.rs
cbrit: Proto enum variant thing

integration_tests/cork_test.go
```diff
@@ -633,6 +624,148 @@ func (s *IntegrationTestSuite) TestCellarV2() {
...
+													Adaptor: adaptorContract.Hex(),
+													CallData: &steward_proto.AdaptorCall_SwapWithUniswapV1Calls{
+														SwapWithUniswapV1Calls: &steward_proto.SwapWithUniswapAdaptorV1Calls{
+															Calls: []*steward_proto.SwapWithUniswapAdaptorV1{
+																{
+																	Function: &steward_proto.SwapWithUniswapAdaptorV1_SwapWithUniV3_{
+																		SwapWithUniV3: &steward_proto.SwapWithUniswapAdaptorV1_SwapWithUniV3{
+																			Path:         []string{"0x1111111111111111111111111111111111111111", "0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"},
+																			PoolFees:     []uint32{1000, 2000},
+																			Amount:       "3",
```
EricBolten: It seems like we're using the `Amount` and `AmountOutMin` to differentiate calls, and the two here have 3 and 4, but the one in the single function test above also uses 3 with the other values being the same. Perhaps we should make them all different to avoid reading the same one in the event feed when checking?

integration_tests/cork_test.go
```diff
@@ -633,6 +624,148 @@ func (s *IntegrationTestSuite) TestCellarV2() {
...
+			}
+
+			s.T().Logf("got %v logs: %v", len(logs), logs)
+			if len(logs) > 0 {
+				for _, log := range logs {
+					if len(log.Data) > 0 {
+						var event AdaptorSwapWithUniV3
+						err := adaptor_abi.UnpackIntoInterface(&event, "SwapWithUniV3", log.Data)
+						s.Require().NoError(err, "failed to unpack SwapWithUniV3 event from log data")
+						s.Require().Equal(big.NewInt(4), event.AmountOutMin)
```
EricBolten: We've added two swap calls in the Multicall but we're only looking for one event. Also, we're only checking `AmountOutMin`, should we verify the other fields of the event?

proto/adaptors/fees_and_reserves.proto
```diff
@@ -0,0 +1,118 @@
...
+        SetupMetaData setup_meta_data = 6;
+        // Represents function `addAssetsToReserves(uint256 amount)`
+        AddAssetsToReserves add_assets_to_reserves = 7;
+        // Represents function `withdrawAssetsFromReserves(uint256 amount)`
+        WithdrawAssetsFromReserves withdraw_assets_from_reserves = 8;
+        // Represents function `prepareFees(uint256 amount)`
+        PrepareFees prepare_fees = 9;
+    }
+
+    /*
```
EricBolten, line 42: Comments still empty here.

steward/src/cellars/cellar_v2_2.rs
```diff
@@ -0,0 +1,269 @@
...
+                .collect::<Result<Vec<Vec<u8>>, Error>>()?
+                .iter()
+                .for_each(|f| multicall.data.push(Bytes::from(f.clone())));
+
+            Ok(multicall.encode())
+        }
+    }
+}
+
+pub fn get_encoded_function(call: FunctionCall, cellar_id: String) -> Result<Vec<u8>, Error> {
```
EricBolten, line 45: Does 2.2 need setupAdaptor?

EricBolten (commented):
Some more follow-ups.

steward/src/cellars/cellar_v2_2.rs
cbrit: They said no in our call.

EricBolten (approved):
LGTM

PR #201, opened 2023-04-12
Enable aave collateral adaptor

cbrit:


proto/adaptors/aave/aave_v2_enable_asset_as_collateral_adaptor.proto
```diff
@@ -0,0 +1,41 @@
...
+
+import "adaptors/base.proto";
+
+// Represents call data for the Aave AToken adaptor, used to manage lending positions on Aave
+message AaveV2EnableAssetAsCollateralAdaptorV1 {
+    oneof function {
+        /***** BASE ADAPTOR FUNCTIONS *****/
+
+        // Represents function `revokeApproval(ERC20 asset, address spender)`
+        RevokeApproval revoke_approval = 1;
```
EricBolten, line 18: Is this actually necessary for this adapter?

proto/cellar_v2.proto
```diff
@@ -373,5 +374,7 @@ message AdaptorCall {
         CellarAdaptorV1Calls cellar_v1_calls = 15;
         // Represents function calls to the UniswapV3Adaptor V2
         UniswapV3AdaptorV2Calls uniswap_v3_v2_calls = 16;
+        // Represents function calls to the AaveV2EnableAssetAsCollatorAdaptor V1
+        AaveV2EnableAssetAsCollateralAdaptorV1Calls aave_v2_enable_asset_as_collateral_v1_calls = 17;
```
EricBolten, line 389: Looks like we're adding this to the 2.0 proto but have not done 2.2.

steward/src/cellars/adaptors/aave_v2.rs
```diff
@@ -309,3 +311,41 @@ pub(crate) fn aave_debt_token_adaptor_v2_calls(
 
     Ok(calls)
 }
+
+pub(crate) fn aave_v2_enable_asset_as_collateral_adaptor_v1_calls(
```
EricBolten: Like the Uniswap case, this is a separate adaptor so we may want a separate file.

EricBolten (commented):
A few comments.

proto/adaptors/aave/aave_v2_enable_asset_as_collateral_adaptor.proto
cbrit: It's a base function for all protos that derive from BaseAdaptor.sol so I do it by default

proto/cellar_v2.proto
cbrit: AdaptorCall is a top level type in the proto, this applies to both. It's split out in the handler files.

EricBolten: Ah, I see that now, 2.2 is split for some files but not here, roger that.

EricBolten: Though the changes you've made support this adaptor for 2.0 as well, wouldn't we need to add it to the Cellar V2.0 section of this file?

EricBolten: Never mind, I see it now.

EricBolten (approved):
LGTM

PR #204, opened 2023-05-01
Status endpoint and AddPositionToCatalogue support

cbrit:


steward/src/cellars.rs
```diff
@@ -21,30 +21,9 @@ pub const BLOCKED_ADAPTORS: [&str; 3] = [
...
-    // FeesAndReservesAdaptor v1
-    "647d264d800a2461e594796af61a39b7735d8933",
-    // CTokenAdaptor v2
-    "9a384df333588428843d128120becd72434ec078",
-];
-pub const ALLOWED_CATALOGUE_ADAPTORS: [&str; 1] = [
-    // AaveV2EnableAssetAsCollateralAdaptor v1
-    "724feb5819d1717aec5adbc0974a655a498b2614",
-];
+pub const ALLOWED_SETUP_ADAPTORS: [&str; 0] = [];
```
EricBolten, line 24: I suppose we can leave these two here with empty lists in case we need to do it again pre-v7, but if these lists are empty we should be able to delete it from the API entirely.

proto/steward.proto
```diff
@@ -19,6 +19,7 @@ import "cellar_v2.proto";
 service ContractCall {
     // Handles simple contract call submission
     rpc Submit(SubmitRequest) returns (SubmitResponse) {}
+    rpc Status(StatusRequest) returns (StatusResponse) {}
```
EricBolten: In retrospect it's kinda weird that the `Status` call is in the `ContractCall` service but I don't know if it's worth changing.

EricBolten (approved):
Small non-blocking comments, approving the PR but we'll want to wait for governance approval before merging.

josephterrigno (approved):

EricBolten (approved):
LGTM

PR #209, opened 2023-06-26
feat:pubsub support for contract call auth 

cbrit:


cbrit:
Need an updated somm_proto crate merged in the sommelier repo

.gitignore
```diff
@@ -16,3 +16,5 @@ package-lock.json
 
 # editors
 .vscode
+
+dev_vendor/
```
EricBolten: Just a note to remove this once we have updated `somm_proto`.

go.mod
```diff
@@ -23,4 +23,6 @@ replace github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alp
 
 replace github.com/peggyjv/steward/steward_proto_go/steward_proto => ./steward_proto_go/steward_proto
 
+replace github.com/peggyjv/sommelier/v4 => ../sommelier
```
EricBolten: Also noting here we need to update this.

integration_tests/setup_test.go
```diff
@@ -385,13 +387,47 @@ func (s *IntegrationTestSuite) initGenesis() {
...
+			Domain:  "localhost",
+			CaCert:  string(caCert),
+		},
+	}
+	pubsubGenState.PublisherIntents = []*pubsubtypes.PublisherIntent{
+		{
+			SubscriptionId:     aaveCellar.Hex(),
+			PublisherDomain:    "localhost",
+			Method:             0,
+			PullUrl:            "https://localhost/somm1p78uz3z9sgav0m325y0tmzzx3a56h583t3x7cx/cacert.pem",
```
EricBolten: We're using `PUSH` subscription methods. `PullUrl` would be an address to query for cork updates from a strategist. You're thinking of `ProofUrl` from the gov prop for new publishers. There should be no `PullUrl` for a `PUSH` intent.

src/config.rs
```diff
@@ -152,6 +152,21 @@ impl Default for CorkConfig {
...
+#[serde(default)]
+pub struct PubsubConfig {
+    /// Cache refresh period in seconds
+    pub cache_refresh_period: u64,
+}
+
+impl Default for PubsubConfig {
+    fn default() -> Self {
+        Self {
+            cache_refresh_period: 3600,
```
EricBolten, line 175: Since we are examining every gov prop, should we force a cache refresh when a publisher or default subscription proposal passes? Should we restart the server thread as well when we detect a change in the publisher cache?

src/cork.rs
```diff
@@ -68,21 +74,35 @@ impl proto::contract_call_service_server::ContractCallService for CorkHandler {
...
 
-        if let Err(err) = schedule_cork(&cellar_id, encoded_call.clone(), height).await {
-            error!("failed to schedule cork for cellar {}: {}", cellar_id, err);
-            return Err(Status::new(
-                Code::Internal,
-                format!("failed to send cork to sommelier: {}", err),
-            ));
+        match schedule_cork(&cellar_id, encoded_call.clone(), height).await {
+            Ok(res) => {
+                if res.code != 0 {
```
EricBolten, line 67: What type of code is this? If it's an HTTP response code you'd probably expect a 200 response for OK.

src/error.rs
```diff
@@ -66,6 +67,18 @@ pub enum ErrorKind {
...
+    #[error("subscriber not found")]
+    SubscriberNotFound,
+    #[error("publisher not found")]
+    PublisherNotFound,
+    #[error("invalid argument")]
+    InvalidArgument,
+    #[error("no subscriptions")]
+    NoSubscriptions,
+    #[error("parsing error")]
+    ParsingError,
```
EricBolten, line 79: We may want to make this more specific if it's only for X509 certificates.

steward_proto_go/steward_proto/go.mod
```diff
@@ -0,0 +1,16 @@
+module github.com/peggyjv/steward/steward_proto_go/steward_proto
```
EricBolten: What's causing the introduction of a new `go.mod` for `steward_proto_go`? How did it work before without it?

src/pubsub.rs
```diff
@@ -0,0 +1,154 @@
...
+                // if there is an entry for the key, push to the vec and dedup,
+                // otherwise create a new vec with the subscription id
+                .entry(s.publisher_domain.to_owned())
+                .and_modify(|e| {
+                    e.push(s.subscription_id.to_owned());
+                    e.dedup();
+                })
+                .or_insert(vec![s.subscription_id.to_owned()]);
+        });
+    get_subscriber_intents_by_subscriber_address(DELEGATE_ADDRESS.to_string())
```
EricBolten: I think the net sum of these two queries is the union operation rather than overwriting. This would only potentially add additional subscription data to the existing default subscriptions rather than overwriting it.

src/pubsub/cache.rs
```diff
@@ -0,0 +1,99 @@
...
+    subject_key_identifier: &Vec<u8>,
+) -> Option<PublisherTrustData> {
+    let cache = PUBLISHER_TRUST_STATE_CACHE.read().await;
+    cache.get(subject_key_identifier).cloned()
+}
+
+/// Overwrites the cache with the latest queried publisher subject key identifier/trust data pairs
+pub async fn refresh_publisher_trust_state_cache() -> Result<bool, Error> {
+    debug!("refreshing publisher trust state cache");
+    let mut cache = PUBLISHER_TRUST_STATE_CACHE.write().await;
```
EricBolten, line 47: Are the write locks automatically released when the function scope exits?

src/server.rs
```diff
@@ -12,48 +29,207 @@ pub struct ServerConfig {
...
+    let key = tokio::fs::read(&config.server.server_key_path).await?;
+    let cert: Vec<rustls::Certificate> = rustls_pemfile::certs(&mut cert.as_slice())
+        .unwrap()
+        .iter()
+        .map(|v| rustls::Certificate(v.clone()))
+        .collect();
+    let key = match rustls_pemfile::read_one(&mut key.as_slice())
+        .expect("cannot parse private key .pem file")
+    {
+        Some(rustls_pemfile::Item::RSAKey(key)) => rustls::PrivateKey(key),
```
EricBolten: I don't think we need to support RSA keys, we expect everyone to use EC keys. Is this a case of having to match every possible outcome? We might want to return an error.

src/server.rs
```diff
@@ -12,48 +29,207 @@ pub struct ServerConfig {
...
+/// A client is authorized to publish to a subscription if the following criteria are met:
+/// 1. The client certificate has an authority key identifier extension whose value is the key of an entry
+///    in the publisher trust state cache
+/// 2. The client certificate is signed by the publisher's CA certificate
+/// 3. The entry's list of subscription IDs contains the subscription ID of the request
+pub(crate) async fn handle_authorization(
+    subscription_id: &String,
+    certs: Arc<Vec<Certificate>>,
+) -> Result<(), Status> {
+    // In case the request contains multiple certificates, we iterate until the first valid one.
```
EricBolten, line 152: If peer certs are sending multiple certificates, it may be the case that it's providing the entire trust chain, in which case we need to inspect the client cert specifically.

src/server.rs
```diff
@@ -12,48 +29,207 @@ pub struct ServerConfig {
...
+    for cert in certs.iter() {
+        let (_, client_cert) = match X509Certificate::from_der(cert.get_ref()) {
+            Ok(cert) => cert,
+            Err(e) => {
+                error!("failed to parse certificate for authorization check: {e}");
+                continue;
```
EricBolten, line 158: None of the error conditions in `handle_authorization` return an error result, they all `continue` to the empty `Ok` at the end.

src/server.rs
```diff
@@ -12,48 +29,207 @@ pub struct ServerConfig {
...
+        .to_vec();
+
+        let publisher_trust_data = match lookup_trust_data_by_subject_key_identifier(
+            &client_cert_aki,
+        )
+        .await
+        {
+            Some(p) => p,
+            None => {
+                debug!("no publisher trust data found for authority key identifier: {client_cert_aki:?}");
```
EricBolten: Most of the `debug!` calls in here probably should be `error!` calls. Valid connections but errors here means a real publisher is having issues of some kind.

src/server.rs
```diff
@@ -12,48 +29,207 @@ pub struct ServerConfig {
...
+        if let Err(e) = client_cert.verify_signature(Some(ca_pub)) {
+            debug!("failed to verify signature for certificate: {e}");
+            continue;
+        }
+
+        if !publisher_trust_data
+            .subscription_ids
+            .contains(subscription_id)
+        {
+            debug!("publisher has no intent to publish to subscription {subscription_id}");
```
EricBolten: This is a little wonky: technically what you've collected is `SusbcriberIntents` when you gather publisher trust data. The message isn't quite correct here, but IIRC deleting a `PublisherIntent` will delete `SubscriberIntents` for that, so it's roundabout correct but doesn't represent what's being queried here.

src/simulate.rs
```diff
@@ -60,3 +68,49 @@ impl proto::simulate_contract_call_service_server::SimulateContractCallService f
 pub fn get_string_encoded_call(request: ScheduleRequest) -> Result<String, Error> {
     get_encoded_call(request).map(|b| bytes_to_hex_str(&b))
 }
+
+pub fn validate_simulate_tls_config(config: &StewardConfig) {
+    if config.simulate.server_cert_path.is_empty() {
+        panic!("Simulate server cert key is not set");
```
EricBolten, line 74: Any particular reason we went with `panic!` here rather than an error flow?

src/somm_send.rs
```diff
@@ -28,10 +28,10 @@ pub async fn schedule_cork(
         block_height,
     };
     let msg = Msg::new("/cork.v2.MsgScheduleCorkRequest", msg);
-    __send_messages(contact, delegate_key, fee, vec![msg]).await
+    send_messages(contact, delegate_key, fee, vec![msg]).await
 }
 
-async fn __send_messages(
+pub(crate) async fn send_messages(
```
EricBolten: Why are we making `send_messages` effectively public here?

EricBolten (commented):
Some feedback and questions, more to come when we can fix the proto crate and merge conflicts.

src/pubsub/cache.rs
cbrit: yes

src/simulate.rs
cbrit: This and `validate_tenderly_config` are only called from the simulate command handler, which handles config/validation error by either panicking or shutting down the process (need to revisit this second one because it doesn't gracefully shut down) because the `abscissa_core::run` call that wraps that closure they're in doesn't allow graceful error handling

src/config.rs
cbrit: We're already restarting the server thread when the trust state changes, see `start_server_management_thread` in server.rs. 

I can work on refreshing the cache when a prop passes

src/cork.rs
cbrit: When adding the pubsub cli commands, i used it for parsing URLs. I don't see a particular need to having specific ones for each if we aren't making any decisions based on `ErrorKind` for these types of errors yet.

src/error.rs
cbrit: When adding the pubsub cli commands, i used it for parsing URLs. I don't see a particular need to having specific ones for each if we aren't making any decisions based on ErrorKind for these types of errors yet.

src/cork.rs
cbrit: This is an ABCI response code, which I believe is the same as the error code defined in each codespace's error.go file. Tendermint interprets 0 as success according to this spec:
https://github.com/tendermint/spec/blob/master/spec/abci/abci.md#errors

src/server.rs
cbrit: Wasn't thinking. I'm a little confused by the PKCS8 variant; isn't that just a syntax for storing keys in the file? Do we need to determine if the underlying key is RSA or EC if the PKCS8 variant is hit for some reason?

src/pubsub.rs
cbrit: There is a dedup() call on the collection on each entry so it's effectively overwriting. Or do you mean something else?

src/server.rs
cbrit: would I do this by iterating the extension for the `id-kp-clientAuth` object identifier, or can i assume it's the first/last in the chain?

src/server.rs
cbrit: The assumption was if one is valid then it will get all the way to `Ok(())` and we return. If none, the loop gets exited by the last `continue` and the error at the end of the function gets returned. 

I think this structure is going to change given your comment on the need to find the client cert in the chain.

src/server.rs
cbrit: Ah, I suppose it would be more appropriate to say "not subscribed to this publisher for subscription {ID}" or something

steward_proto_go/steward_proto/go.mod
cbrit: Don't remember why I did it. It compiles when I remove it and the `require` line in the root go.mod, like it is in the main branch. Removed

go.mod
```diff
@@ -1,26 +1,204 @@
...
+	nhooyr.io/websocket v1.8.6 // indirect
+	sigs.k8s.io/yaml v1.3.0 // indirect
+)
 
 replace github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 
 replace github.com/peggyjv/steward/steward_proto_go/steward_proto => ./steward_proto_go/steward_proto
 
 replace github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
+
```
EricBolten: This is referring to the line above -- we don't need the ics23 replace directive in this version of the SDK.

integration_tests/setup_test.go
```diff
@@ -386,13 +396,64 @@ func (s *IntegrationTestSuite) initGenesis() {
...
+	pubsubGenState.Publishers = []*pubsubtypes.Publisher{
+		{
+			Address: s.chain.orchestrators[0].address().String(),
+			Domain:  "localhost",
+			CaCert:  string(caCert),
+		},
+	}
+	pubsubGenState.PublisherIntents = []*pubsubtypes.PublisherIntent{
+		{
+			SubscriptionId:     aaveCellar.Hex(),
```
EricBolten: We'll have to account for normalization here -- so these would probably be "1:<addr>".

src/commands/cosmos_to_eth.rs
```diff
@@ -173,9 +173,9 @@ impl Runnable for CosmosToEthCmd {
 
         if !self.wait_for_batch {
             println!("Requesting a batch to push transaction along immediately");
-            send_request_batch_tx(cosmos_key, denom,config.cosmos.gas_price.as_tuple(), &contact,config.cosmos.gas_adjustment)
-                .await
-                .expect("Failed to request batch");
+            // send_request_batch_tx(cosmos_key, denom,config.cosmos.gas_price.as_tuple(), &contact,config.cosmos.gas_adjustment)
```
EricBolten: We should probably just delete it since this doesn't exist any more I don't think.

src/commands/pubsub/add_subscriber.rs
```diff
@@ -0,0 +1,45 @@
...
+        let data = std::fs::read_to_string(&self.ca_path).unwrap_or_else(|e| {
+            status_err!("failed to read CA cert file: {}", e);
+            std::process::exit(1);
+        });
+        validate_ca_cert(data.as_bytes()).unwrap_or_else(|e| {
+            status_err!("invalid CA cert: {}", e);
+            std::process::exit(1);
+        });
+        validate_url(&self.push_url).unwrap_or_else(|e| {
+            status_err!("invalid domain name: {}", e);
```
EricBolten: Do we mean URL rather than domain name here?

src/commands/pubsub/unsubscribe.rs
```diff
@@ -0,0 +1,26 @@
...
+pub struct UnsubscribeCmd {
+    /// The cellar to unsubscribe from
+    #[clap(long, short)]
+    cellar_id: String,
+}
+
+impl Runnable for UnsubscribeCmd {
+    fn run(&self) {
+        abscissa_tokio::run_with_actix(&APP, async {
+            unsubscribe(self.cellar_id.clone()).await.unwrap();
```
EricBolten, line 19: In this case (and others) we're referring to the subscription IDs as addresses, but this also has to be normalized.

src/config.rs
```diff
@@ -152,6 +160,21 @@ impl Default for CorkConfig {
     }
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(default)]
+pub struct PubsubConfig {
+    /// Cache refresh period in seconds
```
EricBolten, line 166: We may want a config option for blocked publishers, so in the event of a key compromise, validators could edit their configs to no longer accept the publisher.

src/config.rs
EricBolten: Bumping this.

src/cork.rs
```diff
@@ -61,21 +62,36 @@ impl proto::contract_call_service_server::ContractCallService for CorkHandler {
...
+                        "failed to schedule cork for cellar {}. code {}, raw log: {}",
+                        cellar_id, res.code, res.raw_log
+                    );
+                    return Err(Status::new(
+                        Code::Internal,
+                        "cork submission failed. this may be a steward configuration problem."
+                            .to_string(),
+                    ));
+                }
+                info!("cork response: {:?}", res)
```
EricBolten: Should this be `debug`?

src/cork/proposals.rs
```diff
@@ -1,452 +1,220 @@
...
         };
-    }
 
-    fn increment_proposal_id(&mut self) {
-        self.last_processed_proposal_id += 1;
+    if cork_proposal.block_height <= state.last_observed_height {
+        info!(
+            "proposal {} block height {} has already passed.",
+            proposal_id, cork_proposal.block_height
+        );
```
EricBolten, line 40: Should we be returning early here rather than proceeding?

src/cork/proposals.rs
```diff
@@ -1,452 +1,220 @@
...
+                    return;
                 }
             }
-            Call::CellarV22(data) => {
-                if data.function.is_none() {
-                    warn!(
-                        "scheduled cork proposal {} call data contains no function data and will be ignored: {:?}",
-                        proposal.proposal_id, data,
+        }
+        Call::CellarV22(data) => {
```
EricBolten, line 132: Do we also need v2.5 encodings for governance calls?

src/proposals.rs
```diff
@@ -0,0 +1,269 @@
...
+
+type GovQueryClient = gravity_bridge::gravity_proto::cosmos_sdk_proto::cosmos::gov::v1beta1::query_client::QueryClient<Channel>;
+type GravityQueryClient =
+    gravity_bridge::gravity_proto::gravity::query_client::QueryClient<Channel>;
+type TendermintQueryClient = gravity_bridge::gravity_proto::cosmos_sdk_proto::cosmos::base::tendermint::v1beta1::service_client::ServiceClient<Channel>;
+
+// Proposal types
+const ADD_PUBLISHER_PROPOSAL: &str = "/pubsub.v1.AddPublisherProposal";
+const REMOVE_PUBLISHER_PROPOSAL: &str = "/pubsub.v1.RemovePublisherProposal";
+// const ADD_DEFAULT_SUBSCRIPTION_PROPOSAL: &str = "/pubsub.v1.AddDefaultSubscriptionProposal";
```
EricBolten: I think we may want to add these back (and below) so that we do refresh the cache when a new default subscription is added.

src/proposals.rs
```diff
@@ -0,0 +1,269 @@
...
+                                )));
+                            }
+                        }
+
+                        found_proposal = true;
+                        break;
+                    }
+
+                    if found_proposal {
+                        state.last_processed_proposal_id += missing_proposals;
```
EricBolten, line 155: We're counting missing proposals, but when we find one, won't that have the ID? Couldn't we read it directly?

src/pubsub.rs
```diff
@@ -0,0 +1,215 @@
...
+    let delegate_address = get_delegate_address().to_string();
+    get_subscriber_intents_by_subscriber_address(delegate_address)
+        .await?
+        .subscriber_intents
+        .iter()
+        .for_each(|si| {
+            subscription_mappings
+                .entry(si.publisher_domain.to_owned())
+                .and_modify(|e| {
+                    e.push(si.subscription_id.to_owned());
```
EricBolten: I think the issue here is that subscriber intents must override default subscriptions. Which means that first collecting the default subscriptions and push/dedup, and then doing the same for the subscriber intents on the same map, would have the effect of duplicating a subscription to two different publishers if the subscriber had overridden the default value.

src/server.rs
```diff
@@ -12,48 +29,211 @@ pub struct ServerConfig {
...
-        config.server.client_ca_cert_path.clone(),
-    )
-    .await?;
+pub(crate) async fn start_server_management_thread(mut rx: Receiver<()>) {
+    // wait for the cache to be initialized
+    rx.recv().await;
+
+    loop {
+        let cancellation_token = CancellationToken::new();
+        let server_handle = tokio::spawn(start_server(cancellation_token.clone()));
```
EricBolten, line 38: I assume you tested this, but does the cancellation token work if you clone it rather than pass a reference to the existing one?

src/server.rs
EricBolten: Bumping this, I am not sure what assumptions you can make without inspecting the output.

src/server.rs
EricBolten: We'll come back to this after resolving the other comment.

EricBolten (commented):
Some comments and questions.

src/commands/pubsub/unsubscribe.rs
cbrit: I can change the underlying handlers to normalize them, and add chain ID to the command. To your first point I don't think I understand what you're saying; change the arg name?

src/config.rs
cbrit: > Bumping this.

I believe this is accomplished here, but for some reason I commented out default subscription types. I'll uncomment them.

https://github.com/PeggyJV/steward/blob/bace61740eb801ea4a9ac99052356b25f8a57308/src/proposals.rs#L204

src/cork/proposals.rs
cbrit: will handle in the parity PR

src/proposals.rs
cbrit: It's not actually processing the found one, it's setting the ID to the one right before the one that was found so that the next loop can process it.

src/server.rs
cbrit: Yes, it's designed to cancel all clones of the original when one is cancelled
https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html#method.clone

src/pubsub.rs
cbrit: Ah good catch.

src/commands/pubsub/unsubscribe.rs
EricBolten: If the commands are composing a cellar ID from a chain ID and a cellar address we should probably be explicit about the names of each element. Calling addresses "cellar IDs" was a side effect of there only being one chain ID involved.

go.mod
```diff
@@ -1,26 +1,202 @@
...
-	github.com/ory/dockertest/v3 v3.8.1
-	github.com/peggyjv/gravity-bridge/module/v2 v2.0.2
-	github.com/peggyjv/sommelier/v4 v4.0.2-0.20230124233240-d214e8c0681b
-	github.com/spf13/viper v1.10.1
-	github.com/stretchr/testify v1.8.0
-	github.com/tendermint/tendermint v0.34.14
-	google.golang.org/grpc v1.45.0
-	google.golang.org/protobuf v1.27.1
+	github.com/ory/dockertest/v3 v3.10.0
+	github.com/peggyjv/gravity-bridge/module/v4 v4.0.0-20231211214131-bf2a31eb986f
```
EricBolten: We can point at the v4.0.0 release now without a commit tag.

src/pubsub.rs
```diff
@@ -0,0 +1,254 @@
...
+                .pubsub
+                .publisher_domain_block_list
+                .contains(&s.publisher_domain)
+            {
+                debug!(
+                    "publisher domain {} is in the block list. skipping.",
+                    s.publisher_domain
+                );
+
+                return;
```
EricBolten: Wouldn't we want the blocklist to be enforced in the authorization handler rather than in the cache? That would have immediate effect, although I suppose a config change requires the process to restart anyway, so I'm not sure there.

In either case, we don't need to process the block list twice, we can create the overall mappings and then filter out blocked entries.

Also one more question: in a for_each like this, is return what we want? I guess it's because it's a code block in the iterator rather than an actual loop?

src/pubsub.rs
```diff
@@ -0,0 +1,254 @@
...
+                    "publisher domain {} is in the block list. skipping.",
+                    si.publisher_domain
+                );
+
+                return;
+            }
+
+            // if there is already a default entry for this subscription ID under
+            // a different publisher, we remove it from that set of IDs.
+            if let Some(p) = subscription_mappings.keys().find(|k| {
```
EricBolten: I think what we're doing here is maybe the reverse order of what we want. We're creating subscription mappings with the publisher as the key, which means we have to do a bunch of weird searching. What if instead the subscription ID is the key, and the publisher is the value? Then it would look like this:

1. Write out default subscription mapping
2. Run through subscriberIntent mapping -- simply overwrite the publisher key, if the subscriber is using the default it won't change, if they've customized it it will

EricBolten (commented):
A few final comments.

src/pubsub.rs
cbrit: Ah I'll move it outside. 

Yeah because it's a closure, return acts like `continue` inside of the for_each method

src/pubsub.rs
cbrit: Yeah that would be simpler

EricBolten (approved):
LGTM

PR #210, opened 2023-06-27
Frax, and Morpho adaptors support

cbrit:
Waiting for audits to complete to add necessary addresses to the allow lists


steward/src/cellars.rs
```diff
@@ -1,48 +1,77 @@
...
+pub const CELLAR_RYENS: &str = "18ea937aba6053bc232d9ae2c42abe7a8a2be440";
+pub const CELLAR_RYETH: &str = "b5b29320d2dde5ba5bafa1ebcd270052070483ec";
+pub const CELLAR_RYLINK: &str = "4068bdd217a45f8f668ef19f1e3a1f043e4c4934";
+pub const CELLAR_RYSNX: &str = "cbf2250f33c4161e18d4a2fa47464520af5216b5";
+pub const CELLAR_RYUNI: &str = "6a6af5393dc23d7e3db28d28ef422db7c40932b6";
+pub const CELLAR_RYUSD: &str = "97e6e0a40a3d02f12d1cec30ebfbae04e37c119e";
 
-    Ok(())
-}
+// deprecated cellars
```
EricBolten: These are adaptors, not cellars. We should change the comment and the `CELLAR_` prefix.

steward/src/cellars.rs
```diff
@@ -1,48 +1,77 @@
...
+
+pub const CELLAR_UNIV3_V1: &str = "7c4262f83e6775d6ff6fe8d9ab268611ed9d13ee";
+pub const CELLAR_VESTING_SIMPLE_V1: &str = "1eaa1a100a460f46a2032f0402bc01fe89faab60";
+pub const CELLAR_COMPOUND_C_TOKEN_V1: &str = "26dba82495f6189dde7648ae88bead46c402f078";
+
+// adaptors
+
+pub const ADAPTOR_CELLAR_V2: &str = "24eeaa1111dac1c0fe0cf3c03bba03adde1e7fe4";
```
EricBolten: This isn't the address provided by the smart contracts team, perhaps the wrong one was copy-pasted?

steward/src/cellars.rs
```diff
@@ -1,48 +1,77 @@
...
+pub const ADAPTOR_MORPHO_AAVE_V2_A_TOKEN_V1: &str = "1a4cb53edb8c65c3df6aa9d88c1ab4cf35312b73";
+pub const ADAPTOR_MORPHO_AAVE_V2_DEBT_TOKEN_V1: &str = "407d5489f201013ee6a6ca20fccb05047c548138";
+pub const ADAPTOR_MORPHO_AAVE_V3_A_TOKEN_V1: &str = "b46e8a03b1aafffb50f281397c57b5b87080363e";
```
EricBolten: Nit: this is the A Token Collateral adaptor rather than just a A Token adaptor.

steward/src/cellars.rs
```diff
@@ -93,6 +200,12 @@ mod tests {
         assert!(BLOCKED_ADAPTORS.contains(&normalize_address(blocked3).as_str()));
         assert!(BLOCKED_ADAPTORS.contains(&normalize_address(blocked4).as_str()));
 
-        assert!(!ALLOWED_SETUP_ADAPTORS.contains(&normalize_address(blocked1).as_str()));
+        assert!(!ALLOWED_SETUP_ADAPTORS
```
EricBolten, line 205: I think it would be wise to add some small unit tests for the other new check/validate functions.

steward/src/cellars.rs
```diff
@@ -61,8 +90,86 @@ pub fn normalize_address(address: String) -> String {
...
+pub fn check_blocked_adaptor(adaptor_id: &str) -> Result<(), Error> {
+    let adaptor_id = normalize_address(adaptor_id.to_string());
+    if BLOCKED_ADAPTORS.contains(&adaptor_id.as_str()) {
+        return Err(sp_call_error(format!("adaptor {adaptor_id} is blocked")));
+    }
+
+    Ok(())
+}
+
+pub fn check_blocked_position(position: u32) -> Result<(), Error> {
```
EricBolten: Nit: should these functions just take a `&position` reference as a parameter rather than turning it into a reference in the `contains` call?

steward/src/cellars.rs
```diff
@@ -61,8 +90,86 @@ pub fn normalize_address(address: String) -> String {
...
+        return Err(sp_call_error(format!("position {position} is blocked")));
+    }
+
+    Ok(())
+}
+
+pub fn validate_add_position(cellar_id: &str, position: u32) -> Result<(), Error> {
+    check_blocked_position(position)?;
+    let cellar_id = normalize_address(cellar_id.to_string());
+    if !ALLOWED_ADD_POSITION.contains(&(&cellar_id, position)) {
```
EricBolten, line 116: Out of curiosity, why are we checking blocked positions using `contains` with a uint reference, but here in these tuples we use the owned `position`? Also, in e.g. the `BLOCKED_ADAPTORS.contains` call, we take the output of `normalize_address` and do `&adaptor_id.as_str()`, whereas here we take the output without calling `as_str()` on it. Is there an implicit conversion happening here, or is this checking for the presence of a String type rather than a str somehow? We can validate these with some unit tests.

steward/src/cellars.rs
```diff
@@ -61,8 +90,86 @@ pub fn normalize_address(address: String) -> String {
...
+            "adaptor {adaptor_id} not allowed to be added to catalogue for {cellar_id}"
+        )));
+    }
+
+    Ok(())
+}
+
+pub fn validate_add_position_to_catalogue(cellar_id: &str, position: u32) -> Result<(), Error> {
+    let cellar_id = normalize_address(cellar_id.to_string());
+    if !ALLOWED_CATALOGUE_POSITIONS.contains(&(&cellar_id, position)) {
```
EricBolten, line 141: Same question here re: position reference.

steward/src/cellars/cellar_v2.rs
```diff
@@ -9,26 +9,24 @@ use steward_proto::steward::{
...
+    validate_setup_adaptor,
+};
 
 const CELLAR_NAME: &str = "CellarV2";
 
 /// Encodes a call to a CellarV2 contract
 pub fn get_encoded_call(function: StrategyFunction, cellar_id: String) -> Result<Vec<u8>, Error> {
     match function {
         AddPosition(params) => {
-            if BLOCKED_POSITIONS.contains(&params.position_id) {
```
EricBolten, line 26: Noting here in the old calls position IDs were passed as references.

steward/src/utils.rs
```diff
@@ -116,6 +116,13 @@ pub fn sp_call_parse_address(address: String) -> Result<H160, Error> {
     }
 }
 
+pub fn sp_call_parse_h256(value: String) -> Result<H256, Error> {
```
EricBolten: Is anything using this or is it left over from Balancer?

steward_abi/src/lib.rs
```diff
@@ -7,12 +7,19 @@ pub mod aave_v2_enable_asset_as_collateral_adaptor_v1;
 pub mod aave_v2_stablecoin;
 pub mod aave_v3_a_token_adaptor_v1;
 pub mod aave_v3_debt_token_adaptor_v1;
+pub mod balancer_pool_adaptor_v1;
```
EricBolten: Balancer cruft here as well. There are also the generated Rust and Go files for Balancer still around. We can either keep the generated files around for later (would probably want to add back the Balancer JSON file to the ABI dir) and wait to integrate them, or we should remove them entirely.

EricBolten (commented):
Some fixes, comments, and questions.

steward/src/cellars.rs
cbrit: a). `_::contains()` takes a `&T`. For `BLOCKED_POSITIONS`, `T` is `u32`, and for `ALLOWED_ADD_POSITION`, `T` is a tuple `(&_, _)`. 

In both collections the `position` value is owned by the collection.

b) This is a qwerk of the same `contains()` signature. Since `as_str()` spits out a `&str`, the contains call wants a `&&str`. The same thing is happening on this line, but the deref of the `String` is inside of the tuple.

For some reason, when a `&str` is required, the compiler is ok with using `&String`, but when a `&&str` is wanted, it will not accept `&&String` or `&String`, thus the `.as_str()` call.

EricBolten (approved):
LGTM

PR #211, opened 2023-06-28
CI bug for main/release

cbrit:


.github/workflows/integration_tests.yml
```diff
@@ -95,14 +94,26 @@ jobs:
...
+          images: ${{ env.IMAGE_NAME }}
       - name: Make upload dir
         run: mkdir -p ${{ env.UPLOAD_DIR }}
       - name: build-and-push
         uses: docker/build-push-action@v2
         with:
           file: Dockerfile
-          push: false
-          tags: steward:prebuilt
+          push: ${{ github.event_name != 'pull_request' }}
```
EricBolten, line 101: I suppose my main question here is: do we need to push to Docker at all for steward and hardhat images in our CI workflows? Couldn't we just build the steward and hardhat docker images as artifacts and remove the push aspect entirely?

EricBolten (commented):
A question around Docker usage.

.github/workflows/integration_tests.yml
cbrit: my thinking was we wanted the `steward:main` tag to get pushed when merging to main?

afa hardhat, maybe not so much

EricBolten: We don't really need to publish any docker images unless something else needs to pull them. Steward as a repo is effectively a leaf node in the dependency tree.

EricBolten (approved):
LGTM

PR #213, opened 2023-07-07
Include 3.x line features

cbrit:


docs/01-Configuration.md
```diff
@@ -39,11 +39,11 @@ Configuration related interactions with the Cosmos chain in question
 
 Type: float
 
-Multiplied by estimated gas fee per transaction. It is recommended to set this to a value above 1.0 to avoid gas estimation issues.
+Multiplied by estimated gas fee per transaction. Currently Sommelier fee requirements are 0.0 so this can be left as default.
```
EricBolten: I think we'd want to revert the changes to this doc.

proto/steward/v4/cellar_v2.proto
```diff
@@ -196,3 +191,290 @@ message CellarV2Governance {
...
+}
+
+// Represents a call to adaptor an. The cellar must be authorized to call the target adaptor.
+message AdaptorCall {
+    // Address of the adaptor
+    string adaptor = 1;
+    // The function call data for the adaptor
+    oneof call_data {
+        // Represents function calls to the AaveATokenAdaptor V1
+        AaveATokenAdaptorV1Calls aave_a_token_v1_calls = 3;
```
EricBolten: Are we also skipping from id 1 to 3 on 3.x-main?

proto/steward/v4/fees_and_reserves.proto
```diff
@@ -0,0 +1,118 @@
...
+*/
+
+syntax = "proto3";
+package steward.v4;
+
+option go_package = "/steward_proto";
+
+import "base.proto";
+
+// TODO: Comments
```
EricBolten: Cruft?

src/cellars.rs
```diff
@@ -1,39 +1,120 @@
...
+pub const BLOCKED_ADAPTORS: [&str; 3] = [
+    ADAPTOR_UNIV3_V1,
+    ADAPTOR_VESTING_SIMPLE_V1,
+    ADAPTOR_COMPOUND_C_TOKEN_V1,
+];
+pub const BLOCKED_POSITIONS: [u32; 9] = [4, 5, 6, 7, 8, 9, 10, 11, 12];
+
+// cellars
+
+pub const CELLAR_RY1INCH: &str = "c7b69e15d86c5c1581dacce3cacaf5b68cd6596f";
```
EricBolten, line 30: This looks like we're missing the allowed catalogue/position updates we just added recently to 3.x-main for RYBTC/RYETH. We also still need the setup adaptor allowances for RYUSD as I do not believe they have been applied yet.

src/cellars/adaptors/aave_v2.rs
```diff
@@ -0,0 +1,315 @@
+use abscissa_core::tracing::log::debug;
+use ethers::{abi::AbiEncode, types::Bytes};
+
+use crate::{
+    abi::adaptors::{
+        aave_v2_a_token_adaptor_v1::AaveATokenAdaptorV1Calls as AbiAaveATokenAdaptorV1Calls,
```
EricBolten, line 6: Is support for v1 of the Aave V2 adaptor a leftover from the original Aave cellar?

src/cellars/cellar_v2_2.rs
```diff
@@ -0,0 +1,368 @@
...
+}
+
+pub fn get_encoded_governance_call(
+    function: GovernanceFunction,
+    cellar_id: &str,
+    proposal_id: u64,
+) -> Result<Vec<u8>, Error> {
+    match function {
+        GovernanceFunction::AddAdaptorToCatalogue(params) => {
+            log_governance_cellar_call(
```
EricBolten, line 159: Our governance functions here and in Cellar V2 should respect blocked adaptors and blocked positions.

src/commands/cork_proposal/cellar_v2_2/add_adaptor_to_catalogue.rs
```diff
@@ -0,0 +1,72 @@
...
+        CellarV22governance, GovernanceCall,
+    },
+};
+
+#[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION\n\nCalls addAdaptorToCatalogue() on the target V2.2 cellar contract at the specified block height.\nFor more information see https://github.com/PeggyJV/cellar-contracts/blob/main/src/base/Cellar.sol"
+)]
+pub struct AddAdaptorToCatalogueCmd {
+    #[clap(short = 'a', long)]
```
EricBolten: Any reason why we call out a specific `short` value here but not for others?

src/commands/cork_proposal/cellar_v2_2/force_position_out.rs
```diff
@@ -0,0 +1,76 @@
...
+    #[clap(short = 'p', long)]
+    /// The ID of the position
+    position_id: u32,
+
+    #[clap(short, long)]
+    /// The index of the position
+    index: u32,
+
+    /// Whether the position is a debt position
+    debt: bool,
```
EricBolten, line 30: Does this need a `clap` macro? Also, the field refers to `in_debt_array`, is this shortened intentionally?

src/config.rs
```diff
@@ -210,6 +211,7 @@ impl Default for CosmosSection {
             key_derivation_path: "m/44'/118'/0'/0/0".to_owned(),
             prefix: "somm".to_owned(),
             msg_batch_size: 5,
+            gas_limit_per_msg: 20_000_000u64,
```
EricBolten: The chain has since executed a params proposal that raised this limit to 60M I believe.

src/cork.rs
```diff
@@ -67,6 +59,7 @@ impl proto::contract_call_service_server::ContractCallService for CorkHandler {
             }
         };
         debug!("cork: {:?}", encoded_call);
+        info!("hex cork: {:?}", hex::encode(&encoded_call));
```
EricBolten: I think this is leftover cruft we wouldn't want at the info level. Presumably this would replace the above debug message anyway?

EricBolten (commented):
Some questions and comments.

src/cellars.rs
cbrit: Those are moved to governance now, so I removed them. Am I missing something? We added the allowlists as a temporary measure until governance could take over IIRC.

src/cellars/adaptors/aave_v2.rs
cbrit: We decided to only remove support for adaptors when they are deprecated because of a bug or security concern. afaik v1 of this is still valid.

src/commands/cork_proposal/cellar_v2_2/add_adaptor_to_catalogue.rs
cbrit: Nope. I think I copied some args structs from other commands and it was leftover.

src/commands/cork_proposal/cellar_v2_2/force_position_out.rs
cbrit: Good catch on the clap thing. I did shorten it intentionally, yes.

docs/01-Configuration.md
cbrit: I wish I had written down why I did this, I feel like it was a bug I noticed. Will undo for now.

proto/steward/v4/cellar_v2.proto
cbrit: Ah this was because I removed the original bugged uniswap adaptor. Should I reserve 2 since it's now deprecated/blocked, or update them all since the proto changes are breaking anyway?

src/cellars/cellar_v2_2.rs
cbrit: Should setupAdaptor check blocked adaptors?

docs/01-Configuration.md
EricBolten: The other elements in this doc are suggesting a gas multiplier of 1.0. The current default is 1.1 to avoid the failing message bug, and we want to make sure we don't suggest 1.0.

proto/steward/v4/cellar_v2.proto
EricBolten: We can modify ID values with a major proto version update.

src/cellars.rs
EricBolten: Yeah, I was thinking the wrong way about this. We may still want to add in references to the new cellars/adaptors though if we're going to keep these lists around, as most of the values are unused.

src/cellars/cellar_v2_2.rs
EricBolten: Yes, we should disallow blocked adaptors from both `setupAdaptor`/`addAdaptorToCatalogue` and `callOnAdaptor`.

EricBolten (approved):
LGTM

PR #216, opened 2023-07-18
Balancer pool adaptor

cbrit:


proto/adaptors/balancer/balancer_pool.proto
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /*
+    * Claim rewards ($BAL) from LP position
+    *
+    * Represents `function claimRewards(address gauge)`
+    */
+    message ClaimRewards {
+        // The gauge to claim rewards from
+        string guage = 1;
```
EricBolten: `guage` -> `gauge`

steward/src/cellars/adaptors/aave_v3.rs
```diff
@@ -250,6 +250,9 @@ fn get_encoded_adaptor_calls(
             MorphoAaveV3DebtTokenV1Calls(params) => {
                 calls.extend(adaptors::morpho::morpho_aave_v3_debt_token_adaptor_v1_calls(params)?)
             }
+            BalancerPoolV1Calls(params) => calls.extend(
```
EricBolten, line 253: Am I misremembering, or is there a second location that needs these encodings for flash loans? Maybe it's for Aave V2?

proto/adaptors/balancer/balancer_pool.proto
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    enum SwapKind {
+        SWAP_KIND_UNSPECIFIED = 0;
+        SWAP_KIND_GIVEN_IN = 1;
+        SWAP_KIND_GIVEN_OUT = 2;
+    }
+
+    // Data for a single swap executed by `swap`. `amount` is either `amountIn` or `amountOut` depending on the `kind` value.
+    message SingleSwap {
```
EricBolten, line 46: Should we include the message expected types for these submessages like we do for the function call comments?

steward/src/cellars/adaptors/balancer_pool.rs
```diff
@@ -0,0 +1,174 @@
...
+                {
+                    return Err(sp_call_error("invalid swap kind".to_string()));
+                }
+
+                let swaps_before_join = p
+                    .swaps_before_join
+                    .into_iter()
+                    .map(|s| {
+                        Ok(steward_abi::balancer_pool_adaptor_v1::SingleSwap {
+                            pool_id: sp_call_parse_h256(s.pool_id)?.into(),
```
EricBolten: This is the first time we've parsed something as an `h256`, is there a reason we're using the hash type here rather than some raw bytes? We're calling `into` so it's not the type the ABI is expecting, right? What is the ABI type?

steward/src/cellars/adaptors/balancer_pool.rs
```diff
@@ -0,0 +1,174 @@
...
+                            kind: (s.kind - 1) as u8,
```
EricBolten: Just confirming, because of zero values we do this, the protocol is expecting either 0 or 1, right?

steward/src/cellars/adaptors/balancer_pool.rs
```diff
@@ -0,0 +1,174 @@
...
+                if !p.swaps_after_exit.is_empty() && p.swaps_after_exit.iter().any(|s| s.kind == 0)
+                {
+                    return Err(sp_call_error("invalid swap kind".to_string()));
+                }
+
+                let swaps_after_exit = p
+                    .swaps_after_exit
+                    .into_iter()
+                    .map(|s| {
+                        Ok(steward_abi::balancer_pool_adaptor_v1::SingleSwap {
```
EricBolten: Since we are encoding `SingleSwap` and `SwapData` a bunch of times, we may want to break this out into its own function.

steward/src/cellars/adaptors/balancer_pool.rs
```diff
@@ -0,0 +1,174 @@
...
+                            .into_iter()
+                            .map(string_to_u256)
+                            .collect::<Result<Vec<_>, Error>>()?,
+                        swap_deadlines: d
+                            .swap_deadlines
+                            .into_iter()
+                            .map(string_to_u256)
+                            .collect::<Result<Vec<_>, Error>>()?,
+                    },
+                    None => return Err(sp_call_error("swap data must be set".to_string())),
```
EricBolten: Aren't swaps optional?

EricBolten (commented):
Some comments and questions.

steward/src/cellars/adaptors/aave_v3.rs
cbrit: This is the second location. Aave V3 debt adaptor has flashloan which requires its own copy of the protos/encoding

cbrit: Really it's third* cellar_v2 and *_v2_2 each have one

proto/adaptors/balancer/balancer_pool.proto
EricBolten: To be clear, I meant in the comment.

steward/src/cellars/adaptors/balancer_pool.rs
cbrit: the ABI calls for a bytes32, which is bound to in rust as a [u8; 32]. This function parses the string as its 32 byte hex representation, and `into()` casts it to the array infallibly.

steward/src/cellars/adaptors/balancer_pool.rs
cbrit: correct

steward/src/cellars/adaptors/balancer_pool.rs
cbrit: we can double check but crispy said it would revert

steward/src/cellars/adaptors/balancer_pool.rs
EricBolten: So I guess we're using h256 for convenience since it's backed by a `[u8; 32]` and has some conversion functions already built in? Just odd because it really has nothing to do with a hash type.

cbrit: hmm yeah. actually we should probably check if they are getting the pool_id as hex and then we could just change the param to a string and this parsing would make more sense.

steward/src/cellars/adaptors/balancer_pool.rs
```diff
@@ -132,8 +134,15 @@ fn convert_single_swap(swaps: Vec<SingleSwap>) -> Result<Vec<AbiSingleSwap>, Err
     swaps
         .into_iter()
         .map(|s| {
+            let pool_id = hex::decode(s.pool_id).map_err(|e|
```
EricBolten: Are there any checks we'd want to make on the incoming pool ID value/format before we decode it?

EricBolten:
Looks like there are a couple CI issues here as well.

EricBolten (approved):
LGTM

PR #219, opened 2023-08-21
Generic permission structure for cellars

EricBolten:
Given that there are architecture differences between what positions/adaptors/etc should be allowed or blocked, write a generic permission structure and apply it to each cellar architecture using common functions.


cbrit (approved):
LGTM

PR #220, opened 2023-09-01
SetSharePriceOracle for turbo sweth

cbrit:


steward/src/cellars/cellar_v2_5.rs
```diff
@@ -260,6 +260,36 @@ pub fn get_encoded_function(call: FunctionCall, cellar_id: String) -> Result<Vec
 
             Ok(CellarV2_5Calls::DecreaseShareSupplyCap(call).encode())
         }
+        Function::SetSharePriceOracle(params) => {
+            let cellar_id_normalized = normalize_address(cellar_id.clone());
+            let oracle_in = normalize_address(params.oracle.clone());
+            let allowed_oracles = [
+                normalize_address("0x72249f0199EACf6230DEF33A31e80CF76de78f67".to_string()),
```
EricBolten: We should probably just define constants for these the way we do for cellars.

steward/src/cellars/cellar_v2_5.rs
```diff
@@ -260,6 +260,36 @@ pub fn get_encoded_function(call: FunctionCall, cellar_id: String) -> Result<Vec
...
+                normalize_address("0xC47278B65443cE71CF47E8455BB343F2DB11B70e".to_string()),
+            ];
+            let allowed_registry_ids = [U256::from(3), U256::from(5)];
+            let registry_id_in = string_to_u256(params.registry_id.clone())?;
+            if !cellar_id_normalized.eq(CELLAR_TURBO_SWETH)
+                || !allowed_oracles.contains(&oracle_in)
+                || !allowed_registry_ids.contains(&registry_id_in)
```
EricBolten: Technically the validation here is a mapping of either [3, 0x722...] or [5, 0xc47....], not any combination of the two.

steward_proto_rust/src/prost/steward.v3.rs
```diff
@@ -2986,6 +2989,19 @@ pub mod cellar_v2_5 {
         #[prost(string, tag = "1")]
         pub new_cap: ::prost::alloc::string::String,
     }
+    ///
+    /// Allows strategist to set the share price oracle contract
+    ///
+    /// Represents function `setSharePriceOracle(uint256, address)`
+    #[derive(serde::Deserialize, serde::Serialize, Clone, PartialEq, ::prost::Message)]
+    pub struct SetSharePriceOracle {
```
EricBolten, line 2997: I feel like I'm missing something here, I see the proto binding for SetSharePriceOracle but where is the generated encoding function in the ABI bindings?

EricBolten (commented):
Some comments and questions.

steward_proto_rust/src/prost/steward.v3.rs
EricBolten: Oh, I guess it was in the base cellar ABI we already used, and we had just never touched it. I thought we had talked about having to use a different source ABI but it's there after all.

cbrit: A previous 2.5 ABI we generated happened to have it

EricBolten (approved):
LGTM

PR #221, opened 2023-09-11
RYETH/BTC allow Aave V3 adaptor for RYETH/BTC, Cellar adaptor for RYBTC, cachePriceRouter function

cbrit:
Note the cellar v2.2 ABI file changed, but the only strategist function affected was `cachePriceRouter` which is why it was necessary. Somehow RYETH was on a newer version of the ABI than we had in the repo.


proto/cellar_v2.proto
```diff
@@ -387,6 +389,20 @@ message CellarV2_2 {
     message RemovePositionFromCatalogue {
         uint32 position_id = 1;
     }
+
+    /*
+    * Updates the cellar to use the latest price router in the registry. 
+    *
+    * Represents function `cachePriceRouter(bool checkTotalAssets, uint16 allowableRange, address, expectedPriceRouter)`
```
EricBolten: Nit: `address, ` -> `address `

steward/src/cellars.rs
```diff
@@ -116,6 +129,8 @@ pub const ADAPTOR_COMPOUND_C_TOKEN_V1: &str = "26dba82495f6189dde7648ae88bead46c
 
 // adaptors
 
+pub const ADAPTOR_AAVE_V3_A_TOKEN_V1: &str = "76cef5606c8b6ba38fe2e3c639e1659afa530b47";
+pub const ADAPTOR_CELLAR_V1: &str = "1e22adf9e63ef8f2a3626841ddddd19683e31068";
```
EricBolten, line 133: Why are we using V1 instead of V2?

steward/src/cellars/cellar_v2_2.rs
```diff
@@ -233,6 +233,20 @@ pub fn get_encoded_function(call: FunctionCall, cellar_id: String) -> Result<Vec
...
+        Function::CachePriceRouter(params) => {
+            validate_cache_price_router(&cellar_id)?;
+            log_cellar_call(
+                CELLAR_NAME,
+                &CachePriceRouterCall::function_name(),
+                &cellar_id,
+            );
+            let call = CachePriceRouterCall {
+                check_total_assets: params.check_total_assets,
+                allowable_range: params.allowable_range as u16,
```
EricBolten, line 249: The proto message includes an expected address, is there a reason it's unset here?

steward/src/cellars.rs
```diff
@@ -204,6 +219,15 @@ pub fn validate_oracle(
     Ok(())
 }
 
+pub fn validate_cache_price_router(cellar_id: &str) -> Result<(), Error> {
```
EricBolten: Sun mentioned we were expecting `true` and `500` for the args. Should we validate against that?

EricBolten (commented):
Some questions and comments.

steward/src/cellars.rs
cbrit: As noted [here](https://docs.google.com/spreadsheets/u/2/d/1Kf_DT0bWC1imvr5fVsgiYZAHQ1IsYqw8KEycwAgVI0E/edit#gid=0) this change is for the "legacy" cellar adaptor

steward/src/cellars/cellar_v2_2.rs
cbrit: Ah need to fix the proto. I made it match the 2.5 cellar abi by mistake.

steward/src/cellars.rs
cbrit: I can, do we usually validate non-identity arguments?

EricBolten: I'm basing it off of their requests. I assume this is a one-off and not something that is meant to be called much in the future, but you could verify with them.

EricBolten (approved):
LGTM

PR #223, opened 2023-09-19
feat: Frax Debt/Collateral adaptors + permission updates

cbrit:


proto/adaptors/frax/collateral_f_token.proto
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /*
+    * Allows strategists to add collateral to the respective cellar position on FraxLend, enabling borrowing.
+    *
+    * Represents function `addCollateral(IFToken _fraxlendPair, uint256 _collateralToDeposit)`
+    */
+    message AddCollateral {
+        // The FraxLend pair to add collateral to.
+        string fraxlend_pair = 2;
```
EricBolten: Any reason why the IDs are reversed in this message?

proto/adaptors/frax/collateral_f_token.proto
```diff
@@ -0,0 +1,55 @@
...
+        /***** BASE ADAPTOR FUNCTIONS *****/
+
+        // Represents function `revokeApproval(ERC20 asset, address spender)`
+        RevokeApproval revoke_approval = 1;
+
+        /***** ADAPTOR-SPECIFIC FUNCTIONS *****/
+
+        // Represents function `addCollateral(IFToken _fraxlendPair, uint256 _collateralToDeposit)`
+        AddCollateral add_collateral = 2;
+        // Represents function `removeCollateral(uint256 _collateralAmount, IFToken _fraxlendPair)`
```
EricBolten, line 24: Maybe a dumb question but sanity checking: all the other Fraxlend functions start with the `IFToken`, is this an error or just an anomaly?

proto/adaptors/frax/debt_f_token.proto
```diff
@@ -0,0 +1,68 @@
...
+        // The address of the Frax Pair to repay debt on.
+        string fraxlend_pair = 1;
+        // The amount of the debt token to repay.
+        string debt_token_repay_amount = 2;
+    }
+
+    /*
+    * Allows a strategist to call `addInterest` on a Frax Pair they are using
+    *
+    * Represents `function callAddInterest(IFToken fToken)`
```
EricBolten: The comment on the `oneof` above says the first argument here is `_fraxlendPair` as oppoed to `fToken` here.

steward/src/cellars.rs
```diff
@@ -29,16 +29,22 @@ pub const ORACLE2: (U256, &str) = (
...
 
 // permissions
 
 pub const ALLOWED_V2_0_SETUP_ADAPTORS: [(&str, &str); 1] = [(CELLAR_RYUSD, ADAPTOR_CELLAR_V2)];
-pub const ALLOWED_V2_2_CATALOGUE_ADAPTORS: [(&str, &str); 3] = [
+pub const ALLOWED_V2_2_CATALOGUE_ADAPTORS: [(&str, &str); 5] = [
     (CELLAR_RYETH, ADAPTOR_AAVE_V3_A_TOKEN_V1),
     (CELLAR_RYBTC, ADAPTOR_AAVE_V3_A_TOKEN_V1),
-    (CELLAR_RYBTC, ADAPTOR_CELLAR_V1),
+    (CELLAR_RYBTC, ADAPTOR_LEGACY_CELLAR_V1),
```
EricBolten, line 43: Is this implying then that the currently deployed version, despite allowing RYBTC to add the legacy adaptor, didn't have the correct ABI to call?

steward/src/cellars/adaptors/sommelier.rs
```diff
@@ -55,3 +56,54 @@ pub(crate) fn cellar_adaptor_v1_calls(
...
+                    assets: string_to_u256(p.assets)?,
+                    oracle: sp_call_parse_address(p.oracle)?,
+                };
+                calls.push(
+                    AbiLegacyCellarAdaptorV1Calls::WithdrawFromCellar(call)
+                        .encode()
+                        .into(),
+                )
+            }
+            legacy_cellar_adaptor_v1::Function::RevokeApproval(p) => {
```
EricBolten, line 86: Nit: this is message ID 1 but at the bottom of the list.

steward/src/cellars/adaptors/sommelier.rs
```diff
@@ -55,3 +56,54 @@ pub(crate) fn cellar_adaptor_v1_calls(
...
+                let call = steward_abi::legacy_cellar_adaptor_v1::RevokeApprovalCall {
+                    asset: sp_call_parse_address(p.asset)?,
+                    spender: sp_call_parse_address(p.spender)?,
+                };
+                calls.push(
+                    AbiLegacyCellarAdaptorV1Calls::RevokeApproval(call)
```
EricBolten: Nit: is there a reason these are imported with a rename using the `Abi...` pattern? Our other adaptor files don't do that.

EricBolten (commented):
A few comments and questions.

proto/adaptors/frax/collateral_f_token.proto
cbrit: they just changed the param name in the latest contracts, they are still `address` underneath

steward/src/cellars.rs
cbrit: No, it's the correct ABI I just changed the name to clarify that it's the legacy cellar adaptor. Sanity check from seven seas would be nice though.

steward/src/cellars/adaptors/sommelier.rs
cbrit: It's done elsewhere, but only when the proto definition of the same name is directly imported. It's actually not necessary in this file but go copied that way I guess.

steward/src/cellars/adaptors/sommelier.rs
cbrit: so is the other one 🙃

steward/src/cellars.rs
EricBolten: I meant that the old allowance here was to let RYBTC use the legacy adaptor, but it's only as of this update that that adaptor has the ABI with a third argument.

proto/adaptors/frax/collateral_f_token.proto
EricBolten: I was referring to the order in the function call rather than the `IFToken` type.

steward/src/cellars.rs
cbrit: Oh yeah, unclear to me too. Steward didn't have definitions for it at all until now.

josephterrigno (approved):

EricBolten (approved):
LGTM

PR #225, opened 2023-10-11
Support flashloans in Balancer adaptor API

cbrit:


Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Reference: https://www.lpalmieri.com/posts/fast-rust-docker-builds/
 
-FROM rust:1.63 as cargo-chef-rust
+FROM rust:1.65 as cargo-chef-rust
```
EricBolten, line 3: I looked but I didn't see it elsewhere -- are we defining the version of rust that Steward is meant to be built with somewhere that needs to be updated as well? I may just be forgetting how rustc versioning works.

proto/adaptors/balancer/balancer_pool.proto
```diff
@@ -158,6 +181,82 @@ message BalancerPoolAdaptorV1 {
...
+            // Represents function calls to the MorphoAaveV2DebtToken V1
+            MorphoAaveV2DebtTokenAdaptorV1Calls morpho_aave_v2_debt_token_v1_calls = 20;
+            // Represents function calls to the MorphoAaveV3ATokenCollateral V1
+            MorphoAaveV3ATokenCollateralAdaptorV1Calls morpho_aave_v3_a_token_collateral_v1_calls = 21;
+            // Represents function calls to the MorphoAaveV3ATokenP2P V1
+            MorphoAaveV3ATokenP2PAdaptorV1Calls morpho_aave_v3_a_token_p2p_v1_calls = 22;
+            // Represents function calls to the MorphoAaveV3DebtToken V1
+            MorphoAaveV3DebtTokenAdaptorV1Calls morpho_aave_v3_debt_token_v1_calls = 23;
+            // Represents function calls to the BalancerPoolAdaptor V1
+            BalancerPoolAdaptorV1Calls balancer_pool_v1_calls = 24;
```
EricBolten: Out of curiosity, how does this not create a circular dependency?

steward/src/cellars/adaptors/aave_v3.rs
```diff
@@ -250,9 +250,11 @@ fn get_encoded_adaptor_calls(
             MorphoAaveV3DebtTokenV1Calls(params) => {
                 calls.extend(adaptors::morpho::morpho_aave_v3_debt_token_adaptor_v1_calls(params)?)
             }
+            /*
```
EricBolten: Should include a TODO comment here as well indicating why this is commented out.

proto/adaptors/aave/aave_v3_debt_token.proto
```diff
@@ -153,8 +152,10 @@ message AaveV3DebtTokenAdaptorV1 {
             MorphoAaveV3ATokenP2PAdaptorV1Calls morpho_aave_v3_a_token_p2p_v1_calls = 22;
             // Represents function calls to the MorphoAaveV3DebtToken V1
             MorphoAaveV3DebtTokenAdaptorV1Calls morpho_aave_v3_debt_token_v1_calls = 23;
+            // TODO: Circular dep since balancer has adaptor calls with aave v3 debt token.
+            // Need a more permanent solution.
             // Represents function calls to the BalancerPoolAdaptor V1
-            BalancerPoolAdaptorV1Calls balancer_pool_v1_calls = 24;
+            // BalancerPoolAdaptorV1Calls balancer_pool_v1_calls = 24;
```
EricBolten: I'm not sure how to reconcile semver with this update. I feel like it would be really annoying to have to major version bump because we are making a breaking API change, but it feels like that's technically what's happening. Thoughts?

EricBolten (commented):
Some thoughts and comments.

proto/adaptors/aave/aave_v3_debt_token.proto
EricBolten: Honestly maybe since at this point we only really have one consumer and this is so minor, it doesn't matter.

proto/adaptors/balancer/balancer_pool.proto
EricBolten: Follow-up: I think one fix we could use here would be to essentially create a second adaptor for Aave and Balancer just for the flashloan stuff. So one adaptor with the other calls, one adaptor with the flashloans. Then just not include the flashloan adaptor calls as options for flashloans.

Dockerfile
cbrit: no we aren't. this was necessary because some dependency of cargo-chef was unable to compile against 1.63.

we do define the edition of the rust syntax it uses I think but afaik the compiler version is not enforced at the crate level.

proto/adaptors/balancer/balancer_pool.proto
josephterrigno: BalancerPoolAdaptorV1Calls can be removed as we are not allowed to call it in the Flashloan due to reentrancy

cbrit: 1) because it's defined in the same proto file, but joe just mentioned that it can be removed because of reentrancy

2) yeah in my head that would prevent the dependency loop, will give it a try

proto/adaptors/aave/aave_v3_debt_token.proto
cbrit: I feel the same

proto/adaptors/balancer/balancer_pool.proto
cbrit: this would be true for aave v3's flash loan adaptor calls too right? remove its own calls from the set?

proto/adaptors/aave/aave_v3_debt_token_flash_loan.proto
```diff
@@ -0,0 +1,114 @@
...
+import "adaptors/vesting_simple.proto";
+import "adaptors/sommelier/cellar_adaptor.proto";
+import "adaptors/aave/aave_v2_enable_asset_as_collateral_adaptor.proto";
+import "adaptors/sommelier/legacy_cellar_adaptor.proto";
+import "adaptors/frax/debt_f_token.proto";
+import "adaptors/frax/collateral_f_token.proto";
+
+// Represents call data for the Aave Debt Token adaptor, used for borrowing and repaying debt on Aave.
+message AaveV3DebtTokenAdaptorV1FlashLoan {
+    oneof function {
```
EricBolten: I don't think we need a `oneof` for a single function.

proto/adaptors/balancer/balancer_pool_flash_loan.proto
```diff
@@ -0,0 +1,112 @@
...
+import "adaptors/vesting_simple.proto";
+import "adaptors/sommelier/cellar_adaptor.proto";
+import "adaptors/aave/aave_v2_enable_asset_as_collateral_adaptor.proto";
+import "adaptors/sommelier/legacy_cellar_adaptor.proto";
+import "adaptors/frax/debt_f_token.proto";
+import "adaptors/frax/collateral_f_token.proto";
+
+// Represents flash loan call data for the Balancer Pool adaptor V1, for managing pool positions on Balancer.
+message BalancerPoolAdaptorV1FlashLoan {
+    oneof function {
```
EricBolten: Same here re: `oneof`.

proto/adaptors/aave/aave_v3_debt_token_flash_loan.proto
```diff
@@ -0,0 +1,114 @@
...
+        // Represents function `flashLoan(address[] loanToken, uint256[] loanAmount, bytes params)`
+        FlashLoan flash_loan = 5;
```
EricBolten: There are no other fields in this message so this should probably be ID 1.

proto/adaptors/balancer/balancer_pool_flash_loan.proto
```diff
@@ -0,0 +1,112 @@
...
+        MakeFlashLoan make_flash_loan = 7;
```
EricBolten: Same here re: ID number.

proto/adaptors/balancer/balancer_pool_flash_loan.proto
```diff
@@ -0,0 +1,112 @@
...
+        repeated string tokens = 1;
+
+        // The amounts to flash loan
+        repeated string amounts = 2;
+
+        // The data to flash loan
+        repeated AdaptorCallForBalancerPoolFlashLoan data = 3;
+    }
+
+    message AdaptorCallForBalancerPoolFlashLoan {
```
EricBolten, line 50: There is a comment explaining why we've created duplicate adaptor call functions in the Aave V3 Debt Token adaptor, it might be instructive to include the same explanation here.

steward/src/cellars/adaptors/aave_v3.rs
```diff
@@ -171,9 +153,42 @@ pub(crate) fn aave_v3_debt_token_adaptor_v1_calls(
...
+pub(crate) fn aave_v3_debt_token_adaptor_v1_flash_loan_calls(
+    params: AaveV3DebtTokenAdaptorV1FlashLoanCalls,
+) -> Result<Vec<Bytes>, Error> {
+    let mut calls = Vec::new();
+    for c in params.calls {
+        let function = c
+            .function
+            .ok_or_else(|| sp_call_error("function cannot be empty".to_string()))?;
+
+        let aave_v3_debt_token_adaptor_v1_flash_loan::Function::FlashLoan(p) = function;
```
EricBolten: Is this backwards or some kind of weird rust syntax I don't remember? Where does `p` come from, or is this defining it?

EricBolten (commented):
Some follow-up comments and questions.

steward/src/cellars/adaptors/aave_v3.rs
cbrit: it's defining `p`, newish syntax to avoid a match block and keep the var in scope. normally it could be followed by an `else` block but there is only one variant. this will change when i address your other comment about not needing oneof though.

EricBolten (approved):
LGTM

PR #228, opened 2023-10-31
cachePriceRouter call permissions

cbrit:


steward/src/cellars.rs
```diff
@@ -228,14 +236,15 @@ pub fn validate_cache_price_router(
...
-    if !check_total_assets_value || allowable_range_value != 500 {
-        return Err(sp_call_error(
-            "unauthorized arguments for cachePriceRouter call".to_string(),
-        ));
-    }
     let cellar_id_normalized = normalize_address(cellar_id.to_string());
-    if !ALLOWED_CACHE_PRICE_ROUTER.contains(&cellar_id_normalized.as_str()) {
+    let args = (
+        check_total_assets_value,
+        allowable_range_value,
```
EricBolten: The requirements description had these new cache price router calls as accepting `<=500` rather than the previous expectation of exactly 500. This means that the method of specifying exact lists of static values won't map to the acceptable input values.

EricBolten (commented):
Just highlighting one issue to resolve.

steward/src/cellars.rs
```diff
@@ -228,14 +236,16 @@ pub fn validate_cache_price_router(
...
-            "unauthorized arguments for cachePriceRouter call".to_string(),
-        ));
-    }
     let cellar_id_normalized = normalize_address(cellar_id.to_string());
-    if !ALLOWED_CACHE_PRICE_ROUTER.contains(&cellar_id_normalized.as_str()) {
+
+    if !ALLOWED_CACHE_PRICE_ROUTER.iter().any(|(cellar_id, args)| {
+        cellar_id_normalized.eq(cellar_id)
+            && args.0 == check_total_assets_value
+            && args.1 >= allowable_range_value
```
EricBolten: I believe this is meant to be `<=` rather than `>=`.

steward/src/cellars.rs
```diff
@@ -358,10 +368,10 @@ mod tests {
 
         // allows approved cellar/adaptor ID pairs
         let (v2_0_cellar_id, v2_0_approved_adaptor_id) = (CELLAR_RYUSD, ADAPTOR_CELLAR_V2);
-        assert!(
-            validate_new_adaptor(v2_0_cellar_id, v2_0_approved_adaptor_id, &V2_0_PERMISSIONS)
-                .is_ok()
-        );
+        // assert!(
```
EricBolten, line 381: Why is this causing a breakage?

steward/src/cellars.rs
```diff
@@ -228,14 +236,16 @@ pub fn validate_cache_price_router(
...
 ) -> Result<(), Error> {
-    if !check_total_assets_value || allowable_range_value != 500 {
-        return Err(sp_call_error(
-            "unauthorized arguments for cachePriceRouter call".to_string(),
-        ));
-    }
     let cellar_id_normalized = normalize_address(cellar_id.to_string());
-    if !ALLOWED_CACHE_PRICE_ROUTER.contains(&cellar_id_normalized.as_str()) {
+
+    if !ALLOWED_CACHE_PRICE_ROUTER.iter().any(|(cellar_id, args)| {
```
EricBolten: We could probably use a unit test for this validation function.

EricBolten (commented):
Some further fixes and comments.

steward/src/cellars.rs
cbrit: there is no rule for new 2.0 adaptors

steward/src/cellars.rs
cbrit: `args.1` is the upper limit, `allowable_range_value` is the argument. maybe a confusing variable name. the upper limit has to be the same or higher than the passed in value.

steward/src/cellars.rs
```diff
@@ -454,4 +454,15 @@ mod tests {
...
         assert!(validate_oracle(CELLAR_TURBO_SWETH, &ORACLE2.0.to_string(), ORACLE2.1).is_ok());
     }
+
+    #[test]
+    fn test_validate_cache_price_router() {
+        // valid
+        assert!(validate_cache_price_router(CELLAR_TURBO_STETH, true, 400, Some(PRICEROUTER2)).is_ok());
+
+        // invalid
+        assert!(validate_cache_price_router(CELLAR_TURBO_SWETH, true, 500, Some(PRICEROUTER2)).is_err());
```
EricBolten: Shouldn't the 500 values be valid if it's `<=`?

steward/src/cellars.rs
EricBolten: Might make it clearer if we refer to the hardcoded values as permissions rather than args.

EricBolten (commented):
Another followup.

steward/src/cellars/cellar_v2_5.rs
```diff
@@ -274,6 +274,26 @@ pub fn get_encoded_function(call: FunctionCall, cellar_id: String) -> Result<Vec
 
             Ok(CellarV2_5Calls::SetSharePriceOracle(call).encode())
         }
+        Function::CachePriceRouter(params) => {
+            validate_cache_price_router(
+                &cellar_id,
+                params.check_total_assets,
+                params.allowable_range,
+                Some(&params.expected_price_router),
```
EricBolten: We should be normalizing the input we're receiving either here or in the validation function, since we are comparing expected price routers against hardcoded normalized values.

EricBolten (commented):
One more fix.

EricBolten (approved):
LGTM

PR #230, opened 2023-11-01
chore:main feature parity with 3.x-main

cbrit:
- [x] Add latest adaptors
- [x] Add Cellar 2.5
- [x] Update Governance functions
- [x] Update proposal json commands to match governance functions
- [x] Support axelarcork scheduled cork proposals


cbrit:
CI running is blocked on sommelier pointing at a new gravity release.

abi/AuraERC4626AdaptorV1.json
```diff
@@ -0,0 +1,303 @@
+[
```
EricBolten, line 1: I can't seem to comment on the actual file, do you know why there were a lot of deletions from `abi/CellarV2_2.json`? Also the Balancer ABI is much larger than this Aura one, should we delete cruft if there's any?

build.rs
```diff
@@ -26,13 +26,10 @@ fn generate_contract_abis() {
         ("CellarV1", "cellar_v1"),
         ("CellarV2", "cellar_v2"),
         ("CellarV2_2", "cellar_v2_2"),
+        ("UniswapV3AdaptorV1", "uniswap_v3_adaptor_v1"),
         ("UniswapV3AdaptorV2", "uniswap_v3_adaptor_v2"),
         ("AaveATokenAdaptorV1", "aave_a_token_adaptor_v1"),
         ("AaveDebtTokenAdaptorV1", "aave_debt_token_adaptor_v1"),
-        (
```
EricBolten, line 32: Is deleting this safe? I guess if anything were using code it was generating after the regen there'd be a compile error?

go.sum
```diff
@@ -402,8 +402,9 @@ github.com/go-playground/locales v0.13.0 h1:HyWk6mgj5qFqCT5fjGBuRArbVDfE4hi8+e8c
 github.com/go-playground/locales v0.13.0/go.mod h1:taPMhCMXrRLJO55olJkUXHZBHCxTMfnGwq/HNwmWNS8=
 github.com/go-playground/universal-translator v0.17.0 h1:icxd5fm+REJzpZx7ZfpaD876Lmtgy7VtROAbHHXk8no=
 github.com/go-playground/universal-translator v0.17.0/go.mod h1:UkSxE5sNxxRwHyU+Scu5vgOQjsIJAF8j9muTVoKLVtA=
-github.com/go-playground/validator/v10 v10.2.0 h1:KgJ0snyC2R9VXYN2rneOtQcw5aHQB1Vv0sFl1UcHBOY=
 github.com/go-playground/validator/v10 v10.2.0/go.mod h1:uOYAAleCW8F/7oMFd6aG0GOhaH6EGOAJShg8Id5JGkI=
+github.com/go-playground/validator/v10 v10.4.1 h1:pH2c5ADXtd66mxoE0Zm9SUhxE20r7aM3F26W0hOn+GE=
```
EricBolten, line 406: Maybe `go mod tidy`? There are older versions of this still lingering here in the mod file.

integration_tests/proposal_test.go
```diff
@@ -195,3 +196,147 @@ func (s *IntegrationTestSuite) TestScheduledCorkProposal() {
...
+		proposalQueryResponse, _ := govQueryClient.Proposal(context.Background(), &govtypesv1beta1.QueryProposalRequest{ProposalId: proposalID})
+		return govtypesv1beta1.StatusPassed == proposalQueryResponse.Proposal.Status
+	}, time.Second*30, time.Second*5, "proposal was never accepted")
+	s.T().Log("Proposal approved!")
+
+	s.T().Log("Waiting for scheduled cork to be created by steward")
+	axelarcorkQueryClient := axelarcorktypes.NewQueryClient(orchClientCtx)
+	s.Require().Eventually(func() bool {
+		proposalQueryResponse, _ := axelarcorkQueryClient.QueryScheduledCorks(context.Background(), &axelarcorktypes.QueryScheduledCorksRequest{ChainId: chainID})
+		return len(proposalQueryResponse.Corks) > 0
```
EricBolten: There should be a specific number rather than just being greater than zero, right? I believe we'd expect 4?

integration_tests/proposal_test.go
```diff
@@ -195,3 +196,147 @@ func (s *IntegrationTestSuite) TestScheduledCorkProposal() {
...
+		}
+
+		return false
+	}, 3*time.Minute, 10*time.Second, "never reached scheduled height")
+
+    s.T().Log("waiting to see approved cork result")
+    s.Require().Eventually(func() bool {
+        axelarcorkResultQueryResponse, _ := axelarcorkQueryClient.QueryCorkResults(context.Background(), &axelarcorktypes.QueryCorkResultsRequest{ChainId: chainID})
+        if len(axelarcorkResultQueryResponse.CorkResults) > 0 {
+            if !axelarcorkResultQueryResponse.CorkResults[0].Approved {
```
EricBolten, line 335: I guess this is fine in theory, although a more accurate test would check the cork ID against the one we submitted.

proto/steward/v4/aave_v3_debt_token_flash_loan.proto
```diff
@@ -0,0 +1,119 @@
...
+            // Represents function calls to the SwapWithUniswapAdaptorV1
+            SwapWithUniswapAdaptorV1Calls swap_with_uniswap_v1_calls = 11;
+            // Represents function calls to VestingSimpleAdaptor
+            VestingSimpleAdaptorV2Calls vesting_simple_v2_calls = 12;
+            // Represents function calls to the CellarAdaptor
+            CellarAdaptorV1Calls cellar_v1_calls = 13;
+            // Represents function calls to the UniswapV3Adaptor V2
+            UniswapV3AdaptorV2Calls uniswap_v3_v2_calls = 14;
+            // Represents function calls to the AaveV2EnableAssetAsCollatorAdaptor V1
+            AaveV2EnableAssetAsCollateralAdaptorV1Calls aave_v2_enable_asset_as_collateral_v1_calls = 15;
```
EricBolten, line 86: I'm seeing this here, which means maybe we don't want to delete this from the build as mentioned above? I guess it wouldn't cause an error yet because the source files weren't deleted.

proto/steward/v4/cellar_v2.proto
```diff
@@ -426,8 +437,335 @@ message CellarV2_2Governance {
...
+            InitiateShutdown initiate_shutdown = 9;
+            // Represents function `liftShutdown()`
+            LiftShutdown lift_shutdown = 11;
+            // Represents function `removeAdaptorFromCatalogue(address adaptor)`
+            RemoveAdaptorFromCatalogue remove_adaptor_from_catalogue = 14;
+            // Represents function `removePositionFromCatalogue(uint32 positionId)`
+            RemovePositionFromCatalogue remove_position_from_catalogue = 15;
+            // Represents function `decreaseShareSupplyCap(uint192)
+            DecreaseShareSupplyCap decrease_share_supply_cap = 16;
+
```
EricBolten: Nit: extra line.

src/commands/cork_proposal/cellar_v2_5/add_adaptor_to_catalogue.rs
```diff
@@ -0,0 +1,75 @@
...
+
+            let governance_call = GovernanceCall {
+                call: Some(Call::CellarV22(CellarV22governance {
+                    function: Some(Function::AddAdaptorToCatalogue(AddAdaptorToCatalogue {
+                        adaptor: self.adaptor_address.clone(),
+                    })),
+                })),
+            };
+
+            print_proposal(
```
EricBolten: Our `print_proposal` still doesn't account for whether or not this is an Axelar proposal based on the `chain_id` input, as that prop will have different fields from regular cork proposals.

src/commands/cork_proposal/cellar_v2_5/cache_price_router.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// Whether to check total assets
+    check_total_assets: bool,
+
+    /// Allowable range
+    #[clap(short, long)]
+    allowable_range: u32,
+
+    /// Expected price router address
+    #[clap(short, long)]
+    router_address: String,
```
EricBolten: Nit: maybe call this `expected_price_router` to match the call?

src/commands/cork_proposal/cellar_v2_5/force_position_out.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// The ID of the position
+    position_id: u32,
+
+    #[clap(short, long)]
+    /// The index of the position
+    index: u32,
+
+    /// Whether the position is a debt position
+    #[clap(short, long)]
+    debt: bool,
```
EricBolten: Nit: same here re name `debt` vs `in_debt_array`.

src/commands/cork_proposal/cellar_v2_5/set_automation_actions.rs
```diff
@@ -0,0 +1,86 @@
...
+    long_about = "DESCRIPTION\n\nCalls setAutomationActions() on the target V2.5 cellar contract at the specified block height.\nFor more information see https://github.com/PeggyJV/cellar-contracts/blob/main/src/base/Cellar.sol"
+)]
+pub struct SetAutomationActionsCmd {
+    #[clap(short, long)]
+    /// The registry ID
+    registry_id: String,
+
+    /// Actions address
+    #[clap(short, long)]
+    actions_address: String,
```
EricBolten: Nit: same here with the name.

src/commands/cork_proposal/cellar_v2_5/set_share_price_oracle.rs
```diff
@@ -0,0 +1,86 @@
...
+    long_about = "DESCRIPTION\n\nCalls setSharePriceOracle() on the target V2.5 cellar contract at the specified block height.\nFor more information see https://github.com/PeggyJV/cellar-contracts/blob/main/src/base/Cellar.sol"
+)]
+pub struct SetSharePriceOracleCmd {
+    #[clap(short, long)]
+    /// ID of the oracle registry
+    registry_id: String,
+
+    #[clap(short, long)]
+    /// Address of the oracle
+    oracle_address: String,
```
EricBolten: Nit: apologies for the repetition, but just tagging the places where it's different.

src/commands/cork_proposal/cellar_v2_5/set_strategist_platform_cut.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// Fees Distributor subcommand
+#[derive(Command, Debug, Parser)]
+#[clap(
+    long_about = "DESCRIPTION\n\nCalls setStrategistPlatformCut() on the target V2.5 cellar contract at the specified block height.\nFor more information see https://github.com/PeggyJV/cellar-contracts/blob/main/src/base/Cellar.sol"
+)]
+pub struct SetStrategistPlatformCutCmd {
+    #[clap(short, long)]
+    /// New platform cut proportion for the Strategy Provider between 0 and 1e18 representing 0% and 100% respectively.
+    new_platform_cut: u64,
```
EricBolten: Nit: here as well.

src/cork/proposals.rs
```diff
@@ -220,3 +247,255 @@ pub async fn handle_scheduled_cork_proposal(
         }
     }
 }
+
+pub(crate) async fn handle_axelar_scheduled_cork_proposal(
```
EricBolten, line 131: There's a very large amount of repeated code here, when the only differences are the proposal type, the chain ID parameter, and the call we need to make to schedule it. Is there a way we could use one code block and just set some values based on whether a proposal decodes as `ScheduledCork` or `AxelarScheduledCork`?

src/proposals.rs
```diff
@@ -177,10 +186,32 @@ async fn poll_approved_proposals(
...
-            continue;
         }
 
         let proposal = proposal.unwrap();
+        match proposal.status() {
+            ProposalStatus::DepositPeriod | ProposalStatus::VotingPeriod => {
+                info!("proposal {} is in deposit or voting period", proposal_id);
+
+                // return without incrementing so this ID will be checked again after a period
+                break;
```
EricBolten: Is `break` what we meant here? Isn't this supposed to be an infinite loop? Another thing to keep in mind is that a proposal in `DepositPeriod` may not ever enter `VotingPeriod`, since a full deposit isn't necessary to submit a proposal. If we have to halt when we see one, we'd have to wait the entire deposit period potentially before it could execute the next proposal, after which the block height in a successful one may have already passed.

EricBolten (commented):
Some comments and questions.

src/proposals.rs
EricBolten: Hrmm, on further reflection I think given that deposit periods are the same length as voting periods, it wouldn't be the case where it never exits the deposit period -- rather someone waiting the max amount of time in deposit, then voting, which would have the effect of 96 hours of block time, during which another proposal would be stuck queued behind it for up to 48 extra hours.

go.sum
cbrit: doesn't change anything for me

build.rs
cbrit: duplicate entry, see below

cbrit: It was a duplicate entry, see below

abi/AuraERC4626AdaptorV1.json
cbrit: IIRC there was something missing in the ABI, and previously the json included things like bytecode but the updated one doesn't

src/proposals.rs
cbrit: The usage of a `break` was meant to allow the polling thread to sleep for `query_period` before checking the same proposal again, whereas `continue` would check it immediately. It is a little janky since the query period is configurable and could be short.

Can't anyone deposit to a proposal? In the short term we could prevent it through that means; maybe a more robust thread scheduling system for handling proposals in different states is needed?

go.sum
EricBolten: Weird, there must be something else using that as a dependency. Should be fine I think.

src/proposals.rs
EricBolten: I don't think we'd want to get into the business of paying deposits on bad gov props -- it's also a very cheap attack to create a lot of these. The most straightforward solution here in my mind is this:

1. Set the last observed prop ID to the highest value where there are no intermediate states before it (everything is passed/rejected/failed/missing) -- so we'd only increment on observing those states and not increment past if we see anything in deposit/voting period.
2. Reprocess everything after that ID, but keep a hash map of prop IDs we've already submitted to so we don't have to re-do it. They should be idempotent, but without this it will end up with a bunch of TX spam on the chain.
3. We don't need to care about dependency on prior state -- in the end it's up to submitters to time out and schedule when they want these things to execute. Bear in mind that a passed prop cork has an execution height, it will execute then rather than when the proposal is approved.

src/commands/cork_proposal/cellar_v2_5/set_strategist_platform_cut.rs
cbrit: "cut" seems less specific even if it does match?

src/cork/proposals.rs
```diff
@@ -499,3 +264,79 @@ pub(crate) async fn handle_axelar_scheduled_cork_proposal(
...
+            return Err(ErrorKind::CallDecodeError
+                .context(format!(
+                    "failed to decode GovernanceCall JSON {}: {}",
+                    json, err
+                ))
+                .into());
+        }
+    };
+    if governance_call.call.is_none() {
+        warn!(
```
EricBolten: If we only warn on `is_none`, won't the `unwrap` below panic when it continues?

src/cork/proposals.rs
```diff
@@ -74,59 +74,37 @@ pub async fn handle_scheduled_cork_proposal(
...
+            Ok(res) => {
+                debug!("schedule response: {:?}", res);
+                info!(
+                    "governance call for cellar {} scheduled for block {} by proposal {}",
+                    cellar_id, block_height, proposal.proposal_id
+                );
+                break;
+            }
+            Err(schedule_err) => {
+                log_scheduling_status(
```
EricBolten: We've loaded the `confirm` code into `log_scheduling_status`, which makes it seem a bit overloaded considering we're calling it here on an `Err` condition, but it will then attempt to confirm it. I think it's fine to split out the status logging code but we probably want to pull confirm back out.

src/commands/cork_proposal.rs
```diff
@@ -45,10 +44,36 @@ fn print_proposal(height: u64, contract: String, governance_call: GovernanceCall
...
+}
+
+fn get_axelarcork_proposal_json(height: u64, contract: String, governance_call: GovernanceCall) -> String {
+    let json =
+        serde_json::to_string(&governance_call).expect("failed to serialize governance call");
+    let proposal = AxelarScheduledCorkProposal {
+        block_height: height,
+        target_contract_address: contract,
+        contract_call_proto_json: json,
+        ..Default::default()
```
EricBolten, line 68: The primary difference between the regular cork proposal and the Axelar cork proposal is the chain ID field. We're populating it with a default value here rather than pulling it from the command.

src/proposals.rs
```diff
@@ -123,9 +123,33 @@ async fn poll_approved_proposals(
...
+
+    // Whether there are pending proposals to check
+    let mut pending_proposals = false;
     loop {
         // Proposal IDs start at 1, so this should be ok even for the first query after startup.
-        let proposal_id = &state.last_processed_proposal_id + 1;
+        let proposal_id = if pending_proposals {
+            // remove any stale already-processed pending proposal ids from the set in case
+            // they stuck around from a previous pending-processing phase.
+            pending_already_processed.retain(|&x| x > state.last_processed_proposal_id);
```
EricBolten: Given the overall amount of government proposals and how often Steward gets restarted, this should be a really negligible number, it's probably not even necessary to do.

src/proposals.rs
```diff
@@ -161,15 +185,21 @@ async fn poll_approved_proposals(
...
 
                     if found_proposal {
-                        state.last_processed_proposal_id += missing_proposals;
+                        if pending_proposals {
+                            last_checked_pending_id += missing_proposals;
+                        } else {
+                            state.last_processed_proposal_id += missing_proposals;
+                        }
+
+                        continue;
```
EricBolten, line 185: This `continue` addition is new, what's the benefit restarting the loop here brings?

src/proposals.rs
```diff
@@ -161,15 +185,21 @@ async fn poll_approved_proposals(
...
                     } else {
                         info!(
                             "no new proposals. last processed proposal ID: {}",
                             state.last_processed_proposal_id
                         );
-                    }
 
-                    break;
```
EricBolten, line 163: So previously here, we'd always break the loop and wait for the next polling call when any missing props were found?

src/proposals.rs
```diff
@@ -123,9 +123,33 @@ async fn poll_approved_proposals(
...
     let mut gov_client = GovQueryClient::connect(config.cosmos.grpc.clone()).await?;
+
+    // Tracks which pending proposals have been submitted already for skipping
+    let mut pending_already_processed = HashSet::<u64>::default();
+
+    // Used to track what ID to process when there are pending proposals
+    let mut last_checked_pending_id = 0;
+
+    // Whether there are pending proposals to check
+    let mut pending_proposals = false;
```
EricBolten, line 131: I think we've overcomplicated this a bit by adding a state indicating pending proposals. Because you have the `HashSet` above, you can always record when you've submitted any "pending" proposals. The only state we need to increment is the `last_processed_proposal_id`, we just should not increment it passed anything that could be considered pending. If the name no longer makes sense, maybe `last_finalized_proposal_id` makes more sense. We can always re-process pending proposals because the `HashSet` prevents duplication, so we can just always start from the `last_processed_proposal_id` on a new check.

EricBolten (commented):
Some questions, comments, and suggestions.

src/cork/proposals.rs
cbrit: Confirm is only used for making decisions about log messaging. Is the function naming the issue?

src/commands/cork_proposal.rs
cbrit: Whoops good catch

src/proposals.rs
cbrit: coo'

src/proposals.rs
cbrit: Yes we probably should have continued on found and broke on missing but we were always breaking.

src/proposals.rs
cbrit: We want to keep processing if we find a new proposal with ID gaps, and break/wait otherwise. Previously we were always breaking instead of continuing if we found new ones, which imo was a bug.

src/proposals.rs
cbrit: I was imagining a scenario where a later pending ID is submitted before an earlier pending ID, in such a case we wouldn't want to move the last-processed state beyond the ID that hasn't been submitted. This could happen if iD 100 is stuck in deposit period (but will eventually be voted on) and ID 101 is deposited in full and voted on first.

EricBolten: We don't need to care about the prop ID order in that sense -- the way the chain operates is that gov proposals should be fulfilled when they are approved. If someone is sequencing multiple things they want to happen in a certain order they have to account for that either with proposal times or the parameters they are using (e.g. the block execution height).

cbrit: What I meant was steward will stop processing IDs before it, but now that I think about it I don't think that's true because we have the separate last-pending state. Will fix

EricBolten: It's fine to submit a later ID, because we can just inspect the hashset to see if we've already looked at it. We should only increment the prop ID state when it's the next sequential number (or as part of a series of missing prop IDs).

src/cork/proposals.rs
EricBolten: Maybe I misread `confirm`. If it's not inspecting the chain, then I suppose it doesn't matter that much, but seems like it could be simplified.

src/proposals.rs
cbrit: Ok we're in agreement there, I was misreading your original comment. You're saying don't use a separate ID state list for pending proposals, just always start from the state we have, and only increment when the next sequential ID has been processed. Is that right?

EricBolten: Yes, exactly. This way we can "skip" pending states to process completed props, but then we'll end up reprocessing them later to close them out.

src/proposals.rs
```diff
@@ -231,7 +231,7 @@ async fn poll_approved_proposals(
                 info!("ignoring rejected/failed proposal of ID {}", proposal_id);
                 if pending_proposals {
                     pending_skip_count += 1;
-                    pending_already_processed.insert(proposal_id);
+                    pending_finalized.insert(proposal_id);
```
EricBolten: It's a bit below this line, I can't tag it directly, but I think if we run into an "unspecified" proposal status we'll still need to make sure we skip it.

src/proposals.rs
```diff
@@ -176,19 +206,65 @@ async fn poll_approved_proposals(
                 "proposal {} was None even though the query status code indicates it was found.",
                 proposal_id
             );
-            state.increment_proposal_id();
+
+            if pending_proposals {
```
EricBolten: We're repeating this exact block in a lot of places, I wonder if we shouldn't just make a function out of it. You can make the pending proposal data part of the state too if that makes managing it easier, or just have it update mut vars in this function call.

src/cork/proposals.rs
EricBolten: OK, I see what I was confused by here. We call `schedule_cork` and we expect it to go through. On `Ok`, we write down a log message and break. If `schedule_cork` returns an error, then we call `log_scheduling_status`. The first thing that that function does is run `confirm_scheduling`, which runs a query against the chain to see if the cork was scheduled or not. Why are we checking the chain to see if a cork was scheduled when `schedule_cork` failed? Wouldn't `confirm_scheduling` be something we'd call if `schedule_cork` succeeded?

EricBolten (commented):
Some final comments on gov corks.

EricBolten (approved):
LGTM

PR #233, opened 2023-12-14
feat:multichain contract calls

cbrit:


integration_tests/setup_test.go
```diff
@@ -439,15 +439,15 @@ func (s *IntegrationTestSuite) initGenesis() {
 	}
 	pubsubGenState.DefaultSubscriptions = []*pubsubtypes.DefaultSubscription{
 		{
-			SubscriptionId:  aaveCellar.Hex(),
+            SubscriptionId:  fmt.Sprintf("1:%s", aaveCellar.Hex()),
```
EricBolten: Something weird happen with the formatting here?

proto/steward/v4/steward.proto
```diff
@@ -48,11 +48,17 @@ message ScheduleRequest {
         CellarV2 cellar_v2 = 5;
         CellarV2_2 cellar_v2_2 = 6;
     }
+    // The ID of the chain on which the target Cellar resides
+    uint64 chain_id = 7;
```
EricBolten, line 52: I guess we're going to handle adding CellarV2_5 in main parity?

src/commands/cork_proposal/aave_v2_stablecoin/initiate_shutdown.rs
```diff
@@ -37,7 +40,7 @@ pub struct InitiateShutdownCmd {
 impl Runnable for InitiateShutdownCmd {
     fn run(&self) {
         abscissa_tokio::run_with_actix(&APP, async {
-            cellars::validate_cellar_id(&self.cellar_id)
+            cellars::validate_cellar_id(self.chain_id, &self.cellar_id)
```
EricBolten, line 43: Just commenting here since it applies to all of the commands:

`print_proposal` works for `ScheduledCorkProposal`, but `AxelarScheduledCorkProposal` has a chain ID field

src/cork.rs
```diff
@@ -48,10 +58,12 @@ impl proto::contract_call_service_server::ContractCallService for CorkHandler {
         }
 
         debug!("checking publisher authorization for {cellar_id}");
-        handle_authorization(&cellar_id, certs).await?;
+        let subscription_id = format!("{}:{}", chain_id, cellar_id.trim().to_lowercase());
```
EricBolten: These aren't normalized to lowercase, the subscription IDs reflect whatever the output of querying the cellar IDs from the chain is, which after checking is the mixed-case checksum version. If we aren't, we may want to consider reading in the cellar address inputs into Eth types and stringifying them to ensure we get that value.

src/cork.rs
```diff
@@ -62,36 +74,20 @@ impl proto::contract_call_service_server::ContractCallService for CorkHandler {
...
-            }
+        if chain_id == ETHEREUM_CHAIN_ID {
+            handle_cork(&cellar_id, encoded_call.clone(), height).await?;
+        } else {
+            handle_axelar_cork(chain_id, &cellar_id, encoded_call.clone(), height, deadline)
+                .await?;
         }
-        let id = id_hash(height, &cellar_id, encoded_call);
+
+        let id = id_hash(height, &cellar_id, &encoded_call);
```
EricBolten: I checked and currently we have the deadline in the ID hash on `axelarcork`. I think maybe since removed it from the keys we should also remove it from the ID hash, thoughts? Otherwise we'd need a different way to handle it here.

src/cork.rs
```diff
@@ -132,6 +128,70 @@ pub fn get_encoded_call(request: ScheduleRequest) -> Result<Vec<u8>, Error> {
...
+                    cellar_id, res.code, res.raw_log
+                );
+                return Err(Status::new(
+                    Code::Internal,
+                    "cork submission failed. this may be a steward configuration problem."
+                        .to_string(),
+                ));
+            }
+
+            info!("cork response: {:?}", res);
```
EricBolten: Do we mean debug here?

src/cork/client.rs
```diff
@@ -1,31 +1,67 @@
...
+        let cork_result = self.inner_cork.query_cellar_i_ds(cork_request).await?;
+        let axelarcork_request = AxelarQueryCellarIDsRequest {};
+        let axelarcork_result = self
+            .inner_axelar
+            .query_cellar_i_ds(axelarcork_request)
+            .await?;
+
+        let mut result: HashMap<u64, HashSet<String>> = HashMap::new();
+        for id in cork_result.into_inner().cellar_ids {
+            let normalized_id = id.trim().to_lowercase();
```
EricBolten: Same thing here: we don't want to be adjusting the strings we get back from the chain. When they are written to the chain they are converted to Eth addresses that output in the checksum format.

EricBolten (commented):
Some comments and questions.

proto/steward/v4/steward.proto
cbrit: yes

src/cork.rs
cbrit: I say keep it consistent and remove it yeah.

src/cork.rs
```diff
@@ -62,36 +84,20 @@ impl proto::contract_call_service_server::ContractCallService for CorkHandler {
...
-            }
+        if chain_id == ETHEREUM_CHAIN_ID {
+            handle_cork(&cellar_id, encoded_call.clone(), height).await?;
+        } else {
+            handle_axelar_cork(chain_id, &cellar_id, encoded_call.clone(), height, deadline)
+                .await?;
         }
-        let id = id_hash(height, &cellar_id, encoded_call);
+
+        let id = id_hash(height, &cellar_id, &encoded_call);
```
EricBolten: Based on our conversation we've modified the ID hashes of corks in somm to look like `id_hash(height, chain_id, cellar_id, encoded_call)`, so we'll need to update our calculation here as well.

src/cork/client.rs
```diff
@@ -1,31 +1,79 @@
...
+        let cork_result = self.inner_cork.query_cellar_i_ds(cork_request).await?;
+        let axelarcork_request = AxelarQueryCellarIDsRequest {};
+        let axelarcork_result = self
+            .inner_axelar
+            .query_cellar_i_ds(axelarcork_request)
+            .await?;
+
+        let mut result: HashMap<u64, HashSet<String>> = HashMap::new();
+        for id in cork_result.into_inner().cellar_ids {
+            let normalized_id = match to_checksum_address(&id) {
```
EricBolten, line 49: I suppose we can do this for safety's sake, but what we're really concerned with is the data in the incoming request -- the query server on the chain is going to give us strings in the form of checksum addresses.

src/cork/client.rs
```diff
@@ -34,6 +82,20 @@ impl CorkQueryClient {
         id: &str,
     ) -> Result<Response<QueryScheduledCorksByIdResponse>, Status> {
         let request = QueryScheduledCorksByIdRequest { id: id.to_string() };
-        self.inner.query_scheduled_corks_by_id(request).await
+
+        self.inner_cork.query_scheduled_corks_by_id(request).await
+    }
+
+    pub async fn get_axelar_scheduled_corks_by_id(
```
EricBolten, line 89: I saw this and was wondering what called it -- it turns out that I think this PR doesn't include the proposal thread looking at Axelar gov props.

EricBolten (commented):
A couple comments.

src/cork/client.rs
cbrit: ah. I can do it in another PR

EricBolten (approved):
LGTM

PR #241, opened 2024-01-25
feat:Morpho Blue adaptors

cbrit:


steward/src/cellars/adaptors/morpho.rs
```diff
@@ -308,6 +312,203 @@ pub(crate) fn morpho_aave_v3_debt_token_adaptor_v1_calls(
...
+                calls.push(
+                    MorphoBlueDebtAdaptorV1Calls::RevokeApproval(call)
+                        .encode()
+                        .into(),
+                )
+            }
+            morpho_blue_debt_adaptor_v1::Function::BorrowFromMorphoBlue(p) => {
+                let market = p.market.unwrap();
+                let market = steward_abi::morpho_blue_debt_adaptor_v1::MarketParams {
+                    loan_token: sp_call_parse_address(market.loan_token)?,
```
EricBolten, line 410: We might be better off defining a single function for converting a `MarketParams` message from the `market` input since it's copy-pasted everywhere.

steward/src/cellars/adaptors/morpho.rs
```diff
@@ -308,6 +312,203 @@ pub(crate) fn morpho_aave_v3_debt_token_adaptor_v1_calls(
...
+                    spender: sp_call_parse_address(p.spender)?,
+                };
+                calls.push(
+                    MorphoBlueDebtAdaptorV1Calls::RevokeApproval(call)
+                        .encode()
+                        .into(),
+                )
+            }
+            morpho_blue_debt_adaptor_v1::Function::BorrowFromMorphoBlue(p) => {
+                let market = p.market.unwrap();
```
EricBolten: We probably shouldn't be calling `unwrap` and introducing the possibility of a panic. We can have the market params converting function sanity check that the field is present and return an error otherwise, like the other parse functions do.

EricBolten (commented):
Some suggestions.

steward/src/cellars/adaptors/morpho.rs
cbrit: Yeah I tried this, but unfortunately the ABI bindings each create their own version of it

EricBolten: Yeah, I guess we could define one per adaptor, and that would cut the copy/paste in half. We still want to avoid the unwraps though I think.

EricBolten (approved):
LGTM

PR #244, opened 2024-02-02
feat:alternative assets and morpho blue

cbrit:


proto/steward/v4/aave_v3_debt_token.proto
```diff
@@ -7,25 +7,7 @@ package steward.v4;
 
 option go_package = "/steward_proto";
 
-import "a_token.proto";
```
EricBolten, line 10: Were these leftovers from when the flash loan message was in here?

EricBolten (approved):
One question, but LGTM.

proto/steward/v4/aave_v3_debt_token.proto
cbrit: yes

PR #281, opened 2025-03-25
chore: deep_space and ethers update

cbrit:
## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced TLS and secure connection handling for improved certificate validation and logging.
  - Upgraded transaction response processing for clearer feedback during token transfers and contract deployments.
  
- **Refactor**
  - Streamlined key and wallet loading processes for greater reliability.
  - Reorganized module and utility structures to improve maintainability.
  
- **Chores**
  - Updated the build environment with a newer base image to leverage current optimizations and fixes.


src/server.rs
```diff
@@ -58,32 +71,87 @@ pub(crate) async fn start_server(cancellation_token: CancellationToken) {
...
+                    let mut certificates = Vec::new();
+
+                    let conn = tls_acceptor.accept(connection).await.unwrap();
+
+                    let (_, info) = conn.get_ref();
+
+                    if let Some(certs) = info.peer_certificates() {
+                        debug!("found {} peer certificates", certs.len());
+                        for cert in certs {
+                            certificates.push(cert.clone());
```
zmanian, line 122: I have this concern with pushing the peer certificate into a global static because certificates from multiple sessions with different peers connecting get mixed together. 

So like first good peer connects and while good peer is connected evil peer is connected and it treats evil peer as authorized because good peer showed up with a valid certificate. 

I am not 100% sure this will happen in the current code but I don't see anything that really stops this from happening and pattern seems high risk.

cbrit: so this task, in fact the entire loop, runs uniquely each connection, so the `certificates` vec is owned by a single scope that is unique per client call. is there a global static somewhere I'm missing?

zmanian: I see it now. Okay you are correct.

zmanian (approved):

